PR #2, opened 2015-09-09
fixed python version to 3

OlegHahm:


authmillenon:
I generally don't like scripts that enforce a specific version, but given the version cluster fuck that python 3 is, and that these are only auxiliary scripts I let this pass ;)

PR #15, opened 2016-01-06
test-06: added test app for upd single hop test

haukepetersen:
This is a test app for semi-automatic testing of test case 07 (udp-single-hop). not sure of where to put it, so before merging this PR and after the release is out, we should discuss how we should structure these kind of apps/tests?!

### DIFF
--- a/tests/07-single-hop-udp/app/main.c
+++ b/tests/07-single-hop-udp/app/main.c
@@ -1,0 +1,178 @@
+/*
+ * Copyright (C) 2015 Freie Universität Berlin
+ *
+ * This file is subject to the terms and conditions of the GNU Lesser
+ * General Public License v2.1. See the file LICENSE in the top level
+ * directory for more details.
+ */
+
+/**
+ * @ingroup     examples
+ * @{
+ *
+ * @file
+ * @brief       Example application for demonstrating the RIOT network stack
+ *
+ * @author      Hauke Petersen <hauke.petersen@fu-berlin.de>
+ *
+ * @}
+ */
+
+#include <stdio.h>
+
+#include "shell.h"
+#include "msg.h"
+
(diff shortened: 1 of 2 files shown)


haukepetersen:
missnamed the commit, should be `06` instead of `07`

tests/07-single-hop-udp/app/main.c
```diff
@@ -0,0 +1,178 @@
...
+    return NULL;
+}
+
+static int cmd_run(int argc, char **argv)
+{
+    uint32_t last = xtimer_now();
+    ipv6_addr_t addr;
+
+    if (argc != 2) {
+        printf("usage: %s <target IP address>\n", argv[0]);
```
authmillenon: I would prefer, to be able to still set the number, the size and the interval externally. You can provide a pexpect script that pre-set this.

authmillenon:
```python
def get_ll_addr(node):
    node.send("ifconfig\n")
    node.expect(r"inet6 addr: (?P<ll_addr>fe80::[0-9a-f:]+)/64")
    return node.match.group('ll_addr')
```

for node being a `pexpect.spawn` object ;-)

A-Paul:
> [...]  we should discuss how we should structure these kind of apps/tests?!

I had already started a discussion here https://github.com/RIOT-OS/RIOT/issues/3363 regarding unittests. But I think it is generally advisable to structure and classify the tests more, and then autorun and autoreport as much as possible of that boring iterating stuff. :)

authmillenon:
Did not notice this when commenting https://github.com/RIOT-OS/Release-Specs/issues/18#issuecomment-211861250

haukepetersen:
will re-open if still needed...

PR #52, opened 2017-10-27
Task 5.2: Results are very tight margin

miri64:
I think that this is copy-pasta, since the other [6Lo task](https://github.com/RIOT-OS/Release-Specs/blob/master/04-single-hop-6lowpan-icmp/04-single-hop-6lowpan-icmp.md) asks for <10% success rate. Additionally, having always a success rate of <1% in a crowded network with 100 1kB packets (which means at least 11 fragments) seems highly unlikely.

### DIFF
--- a/05-single-hop-route/05-single-hop-route.md
+++ b/05-single-hop-route/05-single-hop-route.md
@@ -33,7 +33,7 @@ different prefix. A static default route has to be used (`fibroute add :: via <d
 
 ### Result
 
-<1% packets lost on the pinging node.
+<10% packets lost on the pinging node.
 
 Task #03
 ========


smlng:
in general such thresholds are highly subjective to actual boards used, as well as there distance, time of test and so on. So even 10% could be hard to achieve under certain conditions - that's what I experienced when doing Task 04.05 and 04.06 (interop: remotes <-> samr21-xpro). I would go as far as to say everything below 50% packet loss is a successful test - at least without further insights on what is done. 

I mean 50% still is an arbitrary choice but it also means packets get through. Harder performance metrics only make sense if we could also ensure stable environment/conditions, or not?

miri64:
>  I would go as far as to say everything below 50% packet loss is a successful test

I would say if we have 50% packet loss something is definitely wrong with the driver and should not be cosidered a successful test. For the at86rf2xx we are running fine with 10%, but in the past, especially if the testbed is crowded I had my fair share of problems with this task.

miri64:
(due to the <1% thing)

miri64:
> why you only change task 2 to 10% and not the other two as well then?

Because the other two are native => are wire bound (or not even that) => should be way more reliable.

miri64:
(or to say it with my initial statement: that would be copy-pasta, just the other way around)

PR #64, opened 2018-06-26
6Lo tests: also check for pktbuf leaks

miri64:
I found some leaks in the packet buffer for certain conditions in
fragmentation. While not fixed in master at the writing of this change
in RIOT's master, this serves as a reminder that that should be
checked.

### DIFF
--- a/05-single-hop-route/05-single-hop-route.md
+++ b/05-single-hop-route/05-single-hop-route.md
@@ -19,6 +19,7 @@ otherwise default routes and address resolution will be auto-configured).
 ### Result
 
 <1% packets lost on the pinging node.
+No leaks in the packet buffer (check with `gnrc_pktbuf_cmd`).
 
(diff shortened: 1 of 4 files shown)


05-single-hop-route/05-single-hop-route.md
```diff
@@ -79,4 +82,4 @@ otherwise default routes and address resolution will be auto-configured ).
 
 ### Result
 
-<1% packets lost on the pinging node.
\ No newline at end of file
+<1% packets lost on the pinging node.
```
cgundogan: what changed in this line?

cgundogan: Alright, thanks for the hint.

miri64:
We should have waited with this if we can fix this until the release, but nevermind ;-).

Issue #62, opened 2018-04-23
Release 2018.04 - RC2


kaspar030:
This issue lists the status of all tests for the [Release Candidate 2 of the 2018.04](https://github.com/RIOT-OS/RIOT/releases/tag/2018.04-RC2) release.

Specs tested:
- [x] [01-ci](https://github.com/RIOT-OS/Release-Specs/blob/1edc542c4e50e1c582d3cd437c6655050eabd6cf/01-ci/01-ci.md)
  - [x] [Task #01](https://github.com/RIOT-OS/Release-Specs/blob/1edc542c4e50e1c582d3cd437c6655050eabd6cf/01-ci/01-ci.md#task-01) @kaspar030 (using murdock result)
  - [x] [Task #02](https://github.com/RIOT-OS/Release-Specs/blob/1edc542c4e50e1c582d3cd437c6655050eabd6cf/01-ci/01-ci.md#task-02) @kaspar030 (using murdock result)
  - [x] [Task #03](https://github.com/RIOT-OS/Release-Specs/blob/1edc542c4e50e1c582d3cd437c6655050eabd6cf/01-ci/01-ci.md#task-03) @kaspar030
  - [x] [Task #04](https://github.com/RIOT-OS/Release-Specs/blob/1edc542c4e50e1c582d3cd437c6655050eabd6cf/01-ci/01-ci.md#task-04) @cladmi [link](https://github.com/RIOT-OS/RIOT/pull/9092#pullrequestreview-118000929)
- [ ] [02-tests](https://github.com/RIOT-OS/Release-Specs/blob/1edc542c4e50e1c582d3cd437c6655050eabd6cf/02-tests/02-tests.md)
  - [ ] [Task #01](https://github.com/RIOT-OS/Release-Specs/blob/1edc542c4e50e1c582d3cd437c6655050eabd6cf/02-tests/02-tests.md#task-01)
  - [ ] [Task #02](https://github.com/RIOT-OS/Release-Specs/blob/1edc542c4e50e1c582d3cd437c6655050eabd6cf/02-tests/02-tests.md#task-02)
  - [x] [Task #03](https://github.com/RIOT-OS/Release-Specs/blob/1edc542c4e50e1c582d3cd437c6655050eabd6cf/02-tests/02-tests.md#task-03)  [run by murdock](https://ci.riot-os.org/RIOT-OS/RIOT/2018.04-branch/latest/output.html)
[...]

PR #66, opened 2018-07-17
07-multi-hop: Fix issues found when doing Release 2018.04

cladmi:
This fixes issues I got when running 07-multi-hop for the last release

My main concern is that `ifconfig` does not print the `/PREFIX` on addresses and I do not know if it is really a wanted thing or not. However it is required for running the tests.

* Update to `nib` shell commands
* Fix nodes not being sorted as `sorted` does not sort in place
* Update to the new `ifconfig` API

And some cleanups

* Update to an unicode pexpect (I think I got decode error but not sure anymore)
* Update to new iotlab- commands to not get deprecation warnings
* Whitespace

Referenced in:

https://github.com/RIOT-OS/Release-Specs/issues/62#issuecomment-388376218

### DIFF
--- a/07-multi-hop/IOTLABHelper.py
+++ b/07-multi-hop/IOTLABHelper.py
@@ -26,7 +26,7 @@ class IOTLABHelper:
         self.testbed = None
 
     def __getLivingNodesFromTestbed(self, site, nodeType):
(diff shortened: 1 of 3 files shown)


cladmi:
Thanks

PR #67, opened 2018-07-20
01-ci: update scripts

cladmi:
Changes are:

* Use python3
* Print execution output
* Remove execution timeout, (I replaced by `subprocess` for some executions).
* use `make test` when possible
* Use `spawnu` for unicode

I had trouble running the tests for 01-ci due to the timeouts.
I took some time to update the scripts to python3 and using make test.

### DIFF
--- a/01-ci/task01.py
+++ b/01-ci/task01.py
@@ -1,9 +1,8 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
-import pexpect
 import os
 import sys
-import time
+import pexpect
 
 if len(sys.argv) < 2:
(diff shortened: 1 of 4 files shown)


01-ci/task01.py
```diff
@@ -1,21 +1,71 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
+
+"""Compile all tests and examples for all boards.
+
+Usage
+-----
+
+```
+usage: task01.py [-h] [--stop] riot_directory
```
miri64: Maybe also add an (optional) `RC`/`tag` argument to make sure the script is on the right RC.

PR #68, opened 2018-07-30
09-compile-and-test-one-board: add script to test board

cladmi:
Add a script to compile and execute test on a board and produce file results.

I want to add a new release spec step where board maintainers run tests on their own board, with their normal working setup and give us the compilation/tests results.

I do not call it a "test" as I do not want it blocking for a release, but more as a goal of gathering information on the current state of board support and test suite.

TODO:

* [x] Readme.md
* [x] Make some board maintainers upload their tests results
* [x] Integrate in the release specs (is there any special thing to do ?)
  * Should be ok
  * [x] Allow title with description in the release manager

### Description

This adds a script that goes into all applications and, if the board is supported, calls `make clean all`, and then `make flash test` if there are tests.

It produces results in `results/board/tests/name/` with `compilation.success` or `.failed` and `test.success` or `test.failed`. The goal was to produce file results with the execution trace on failure. I also kept for tests as analyzing the content could also be useful.

The script is wrote to not re-run compilation and tests if the output file exists.
So if you change the tested release, you should update the result-directory or delete only the tests output you want to re-run.

Making it correctly detect if things changed is a problem that should go in the Makefile and be made reliable there. I want this script to be easy to re-run on failed tests.

### Usage

* Have a clean RIOT repository version using a release version
* Plug one board to your computer and check that you can run tests `make -C tests/events all flash test BOARD=your_board`
* Run the script

       ./compile_and_test_for_board.py path_to_riot_directory board_name [optional_results_directory]
* Write a PR with your results to: https://github.com/cladmi/RIOT_2018_07_tests_results

### Uploading results

For this release, I created a repository on my github account where developers can put the results:

https://github.com/cladmi/RIOT_2018_07_tests_results

Where my first tester uploaded their results already.

### Future work

Begin merging parts of the script to the main riot build system to provide it as an integrated features. I just did not have time to write it before the release and using a PR for testing the release is more complex than an external script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+results/
+__pycache__/
--- a/02-tests/02-tests.md
+++ b/02-tests/02-tests.md
@@ -1,5 +1,25 @@
 ## Goal: Perform test applications on various platforms and check that they run as expected
 
+
+The `./compile_and_test_for_board.py` script can be used to run all compilation
+and **automated** tests for one board, not the manual tests.
+
+### Usage
+
+Run 02 `compile_and_test_for_board.py` script with your test board connected.
+
+    ./compile_and_test_for_board.py path_to_riot_directory board_name [results]
+
+It prints the summary with results files relative to `results_dir/board` to have
+a github friendly output.
+
+Failures and all tests output are saved in files.
+They can be checked with:
+
+    find results/ -name '*.failed'
+    find results/ -name 'test.success'
+
+
 Task #01 - All tests on native
 ==============================
 ### Description
--- a/02-tests/compile_and_test_for_board.py
+++ b/02-tests/compile_and_test_for_board.py
@@ -65,6 +65,7 @@ optional arguments:
 """
 
 import os
+import sys
 import glob
 import shutil
 import logging
@@ -602,7 +603,7 @@ def main():
         print(summary, end='')
     else:
         logger.info('Tests successful')
-    exit(num_errors)
+    sys.exit(num_errors)
 
 
 if __name__ == '__main__':
--- a/02-tests/compile_and_test_for_board.py
+++ b/02-tests/compile_and_test_for_board.py
@@ -406,7 +406,8 @@ class RIOTApplication():
             raise err
         return out
 
-    def make_with_outfile(self, name, args, save_output=False):
+    def make_with_outfile(self, name, args, save_output=False,
+                          setuptasks=None):
         """Run make but save result in an outfile.
 
         It will be saved in `self.resultdir/name.[success|failure]`.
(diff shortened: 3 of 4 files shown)


cladmi:
It can also be used to run tests for [Task #02](https://github.com/RIOT-OS/Release-Specs/blob/master/02-tests/02-tests.md)

cladmi:
> Couldn't this script also be used for Task 02?

That is even why I started using it :) provide tests results for different boards.

The difference from the basic Task 2 is that I also compile everything even if there is no `test` target. I went for this as more "compile time tests" are being added and that getting more compilation outputs could help debug different environments. But that should not be problematic. The specs could also be updated to match this if we really want to.


A thing I may need to adapt if it is for task 2, is that right now, if you re-run it on the same results directory, it does not re-execute compilation and tests for targets that where already successful. Which simplifies a lot handling temporary failures with the hardware and debugging tests issues. But gives "unsafe results" as they may not be up to date but a good enough default for optional tests.

I could change the behavior to by default force re-run, and an option to be `--incremental` instead of being incremental with a `--force` option.

cladmi:
If you want I can move the script to `task 2` and explain here to use the script from there.

miri64:
:+1: to moving the script

miri64:
It's still in 9 though ;-)

cladmi:
I know, I fix some small issues I have when running it.

cladmi:
I squashed the whole history in two commits with one adding the script for task2 and the one adding task 09.

cladmi:
I separated the `flash` task to a specific file output. As it was the problem for one board that it never flashed. Better make it visible.

cladmi:
Note: it will need adapting when https://github.com/RIOT-OS/RIOT/pull/9867 is merged.

It can either wait, or be adapted later.

jia200x:
> Note: it will need adapting when RIOT-OS/RIOT#9867 is merged.
It can either wait, or be adapted later.

It's ok for me to merge and adapt later :)

cladmi:
> It's ok for me to merge and adapt later :)

Ok, just need a ACK and squashing

cladmi:
It will also need to be adapted after https://github.com/RIOT-OS/RIOT/pull/9567 as the `test` detection will change. So maybe wait until both PRs are merged to merge this one.

cladmi:
Updated now that https://github.com/RIOT-OS/RIOT/pull/9567 is merged.

jia200x:
@cladmi are we ready to merge this one?

cladmi:
I think yes, it just needs a positive review and squash.

cladmi:
I noticed I was uploading to another branch name instead of this one… I fixed this.

jia200x:
code style is good (`flake8`) and the script works as described.

cladmi:
It was not squashed :'(

jia200x:
dumb mistake, I owe beers to all maintainers. 
Next steps? :(

jia200x:
sorry :(

cladmi:
I can do a PR that reverts them and add the squashed commit.

PR #75, opened 2018-10-22
Squashed commits of 68

cladmi:
The PR was merged without squashing the commits.
This one reverts all commits and adds squashed versions.

I put the script documentation as commit message.

##### Testing

The current version has the same files as the merge commit in master: `09a90d8`

```
git diff 09a90d8
# says nothing
```

The commit message can be reviewed.

##### Reference PR:

https://github.com/RIOT-OS/Release-Specs/pull/68


jia200x:
Tested, works as expected. Thank you so much!

miri64:
For future reference: I think it doesn't actually make an accidental merge of un-squashed commits more tidy if you revert and then re-apply squashed. Now the history is even more cluttered.

jia200x:
> For future reference: I think it doesn't actually make an accidental merge of un-squashed commits more tidy if you revert and then re-apply squashed. Now the history is even more cluttered.

Ok. Taken

cladmi:
@miri64 my goal to squash was to at least have somewhere the right commit message introducing it. The first commit description was outdated.
Even if the history is a bit more messy, now doing a `git blame` or a `git log -G` points me first to a valid commit and it is what I tried to ensure by doing these reverts.

PR #77, opened 2018-10-26
09-coap: add CoAP release specs

kb2ma:
### Contribution description
Basic coverage for CoAP messaging as described in #73. The goal for this PR is to include non-confirmable and confirmable messaging, block extension, and observe extension. Includes some tests for both gcoap and nanocoap. Includes test of RIOT Resource Directory client implementation (CORD).

Initial commit probably premature, but I wanted to get the message out that I'm working on it for the 2018.10 release. :-)

### Issues/PRs references
Resolves #73

PR #78, opened 2018-10-29
99-compile-and-test-one-board: rename to be the last task

cladmi:
Rename 09-compile-and-test-one-board to 99- to be sure it is the last task.
The goal is to be an optional step that is a bonus to the normal ones.

This will allow adding more easily new mandatory tasks with contiguous numbers.

### Related issues

I noticed the issue in https://github.com/RIOT-OS/Release-Specs/pull/77 as it was being named 10- so after this optional task.


jia200x:
Makes sense to me.

cladmi:
Thanks for the review.

PR #80, opened 2018-11-05
task08: remove outdated remark in lwIP titles

miri64:
Forgot to remove those in #72.


miri64:
> I will need to adapt the issue as well.

Already done that :-)

PR #82, opened 2018-11-06
task06: clarify port choice for 6.2

miri64:
Just noticed this while scanning over the currently ongoing tests: 61616 is not just "another port". It's within the port range of compressible ports for [6LoWPAN's UDP NHC][RFC6282].

[RFC6282]: https://tools.ietf.org/html/rfc6282#section-4.3.1


miri64:
Just talked to him in the hallway, he just did not know.

cladmi:
I knew there were compressible ports I think, but not that this one was chosen because of this :D

PR #85, opened 2018-11-12
07-multi-hop: Add detailed documentation

MrKevinWeiss:
This commit helps clarify individual steps to take to run the test.
It adds documentation on how to use the shell commands.

### DIFF
--- a/07-multi-hop/example_test_guide.md
+++ b/07-multi-hop/example_test_guide.md
@@ -1,0 +1,113 @@
+_The following is a possible way to run the tests, this is just an example_
+## Goal: Check multi-hop connectivity over IPv6
+_We will arrange all 4 nodes in a chain topology:
+node1 <-> node2 <-> node3 <-> node4_
+
+### Pre-setup Steps
+1. Book 4 [iotlab](https://www.iot-lab.info/testbed/dashboard) M3 nodes
+
+# Task 1 and 2 Setup
+1. Open 4 instances of nodes physically close together
+`USEMODULE=gnrc_pktbuf_cmd make -C tests/gnrc_udp/ BOARD=iotlab-m3
+IOTLAB_NODE=<iotlab id> flash term`
+2. On each end of the nodes add a global IP address (ie. node1:`ifconfig 6 add
+    abcd::1` and node 4: `ifconfig 6 add abcd::2`)
+3. Add the routing for both ways:
+- Node 1: `nib route add 6 abcd::2 <ipv6 addr of node 2>`
+- Node 2: `nib route add 6 abcd::2 <ipv6 addr of node 3>` and `nib route add 6
+abcd::1 <ipv6 addr of node 1>`
+- Node 3: `nib route add 6 abcd::2 <ipv6 addr of node 4>` and `nib route add 6
+abcd::1 <ipv6 addr of node 2>`
+- Node 4: `nib route add 6 abcd::1 <ipv6 addr of node 3>`
+3. Proceed to Task specific commands
+
+
+Task #01 - ICMPv6 echo on iotlab-m3 with three hops (static route)
+==================================================================
+### Description
+
+ICMPv6 echo request/reply exchange between two iotlab-m3 nodes over three hops
+with static routes.
+
+1. Follow Task 1 and 2 setup
+2. Ping node 4 from node 1, Node 1:`ping6 100 abcd::2 50 100`
+3. Ping node 1 from node 4, Node 4:`ping6 100 abcd::1 50 100`
+
+### Result
+
+<20% packets lost on the pinging node.
+No leaks in the packet buffer (check with `gnrc_pktbuf_cmd`).
+
+Task #02 - UDP on iotlab-m3 with three hops (static route)
+==========================================================
+### Description
+
+Sending UDP between two iotlab-m3 nodes over three hops with static routes.
+
+1. Follow Task 1 and 2 setup
+2. Start a udp server on node 1, Node 1:`udp server start 1234`
+3. Start a udp server on node 4, Node 4:`udp server start 4321`
+4. Send a packet to node 4 from node 1, Node 1:`udp send abcd::2 4321 50`
+5. Send a packet to node 1 from node 4, Node 4:`udp send abcd::1 1234 50`
+
+### Result
+
+<10% packets lost on the pinging node.
+No leaks in the packet buffer (check with `gnrc_pktbuf_cmd`).
+
+
+# Task 3 and 4 Setup
+1. Flash all nodes with with the l2filter_whitelist module firmware
+`USEMODULE="l2filter_whitelist gnrc_pktbuf_cmd" BOARD=iotlab-m3 IOTLAB_NODE=<iotlab id> make -C tests/gnrc_udp/ flash term`
+2. Setup the l2filter_whitelist addresses
+- Node 1: `ifconfig 6 l2filter add <MAC addr of node 2>`
+- Node 2: `ifconfig 6 l2filter add <MAC addr of node 1>` and `ifconfig 6 l2filter add <MAC addr of node 3>`
+- Node 3: `ifconfig 6 l2filter add <MAC addr of node 2>` and `ifconfig 6 l2filter add <MAC addr of node 4>`
+- Node 4: `ifconfig 6 l2filter add <MAC addr of node 3>`
+3. Add a global IP address, Node 1: `ifconfig 6 add abcd::1/64`
+4. Apply rpl root, Node 1: `rpl root 0 abcd::1`
+5. Check Node 4 for R >= 1024 indicating more at least 3 hops,  Node 4:
+
+```
+> rpl
+rpl
+instance table: [X]
+parent table:   [X]     [ ]     [ ]
+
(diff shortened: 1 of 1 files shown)


07-multi-hop/07-multi-hop.md
```diff
@@ -1,12 +1,38 @@
 ## Goal: Check multi-hop connectivity over IPv6
 
+### Pre-setup Steps
+1. Book n>15 iotlab M3 nodes (preferably in grenoble) in a straight line 3 nodes apart. (so x--x--x--x)
+2. Add the following CFLAGS to the `tests/gnrc_udp`:
+```
+CFLAGS += -DGNRC_PKTBUF_SIZE=512 -DGNRC_NETIF_IPV6_ADDRS_NUMOF=20 \
```
miri64: Why do you make the packet buffer smaller here?

07-multi-hop/07-multi-hop.md
```diff
@@ -1,12 +1,38 @@
...
+          -DGNRC_IPV6_NIB_NUMOF=20 \
+          -DGNRC_IPV6_NIB_OFFL_NUMOF=20
```
miri64: I'm not sure those big numbers are really necessary.

07-multi-hop/07-multi-hop.md
```diff
@@ -1,12 +1,38 @@
 ## Goal: Check multi-hop connectivity over IPv6
 
+### Pre-setup Steps
+1. Book n>15 iotlab M3 nodes (preferably in grenoble) in a straight line 3 nodes apart. (so x--x--x--x)
```
miri64: Why 15? The test specs ask for 5.

07-multi-hop/07-multi-hop.md
MrKevinWeiss: I just took the flags from minimal makefile.  TBH I don't understand everything that goes on., this is just what happened to work (although a lot of instruction was given to me by @cgundogan).  Shall I remove it?

07-multi-hop/07-multi-hop.md
MrKevinWeiss: It works and compiles on the M3 nodes.  I usually book about 20 node and I think if these numbers are too small the routing tables can't handle it.

07-multi-hop/07-multi-hop.md
miri64: I'd argue to keep it as default as possible. Especially for forwarding nodes using `gnrc_minimal` as a template doesn't make much sense, as `gnrc_minimal` is in host configuration and what you need for forwarding is router configuration.

07-multi-hop/07-multi-hop.md
MrKevinWeiss: This is because it is very hard to get proper positioning.  If I book a whole bunch I don't need to alter the position I can just move to the next node.

07-multi-hop/07-multi-hop.md
miri64: But if you got the instructions from @cgundogan I'd like to get his input first.

MrKevinWeiss: OK, I can retest with defaults to see if everything works (after all comments are addressed though)

07-multi-hop/07-multi-hop.md
miri64: Yes of course it works. More address space to allocate doesn't make it suddenly not work ;-). But I'm wondering if it is really necessary to have all that address space.

07-multi-hop/07-multi-hop.md
miri64: From my experience ~7 nodes apart is enough to only have the neighboring nodes reachable. And in the end you still can use `l2filter_blacklist` to remove the nodes that aren't supposed to reach each other.

07-multi-hop/07-multi-hop.md
miri64: You might get problems with RPL and **15** nodes ;-).

miri64: But `GNRC_NETIF_IPV6_ADDRS_NUMOF=20` really isn't required. You just need 2 (global and link-local address) per node.

miri64: Here's the doc for the macros you are configuring here. Based on that you should be able to make a informed decision on the configuration yourself:

- [`GNRC_PKTBUF_SIZE`](http://doc.riot-os.org/group__net__gnrc__pktbuf.html#ga83a14b0588bb8adc93b565c4ebc5ec56)
- [`GNRC_NETIF_IPV6_ADDRS_NUMOF`](http://doc.riot-os.org/netif_2conf_8h.html#a0d2713931d504ecd972747894c3b08ef)
- [`GNRC_IPV6_NIB_NUMOF`](http://doc.riot-os.org/group__net__gnrc__ipv6__nib__conf.html#gad549499055dcce8399ebf30f6cb5073c)
- [`GNRC_IPV6_NIB_OFFL_NUMOF`](http://doc.riot-os.org/group__net__gnrc__ipv6__nib__conf.html#gae319bc32b0670b25529b746b9aa019b1)

MrKevinWeiss:
@miri64 How about I separate this out as a possible way to run it?  Maybe move it to example1.md?  Would that be easier?

miri64:
> @miri64 How about I separate this out as a possible way to run it? Maybe move it to example1.md? Would that be easier?

I mostly agree with what you propose, I just find the choice of configuration questionable for an official release testing procedure: The packet buffer is way too small for routers (if you try to send larger packets you will see that), the amount of configurable addresses way too big. The only thing that is a matter of taste is the number of IoT-LAB nodes picked.

MrKevinWeiss:
I broke it out as a guide (a possible way to test).  I also removed the packet_buf flag as it was added due to my ignorance and reduce the configurable addresses a bit.  I think other people can add there own versions (like adding the blacklisting, I think that would be useful for automated tests).  However, at this time I don't know how that is done.

07-multi-hop/example_test_guide.md
```diff
@@ -0,0 +1,113 @@
+_The following is a possible way to run the tests, this is just an example_
+## Goal: Check multi-hop connectivity over IPv6
+
+### Pre-setup Steps
+1. Book n>15 iotlab M3 nodes (preferably in grenoble) in a straight line 3 nodes apart. (so x--x--x--x)
+2. Add the following CFLAGS to the `tests/gnrc_udp`:
+```
+CFLAGS += -DGNRC_NETIF_IPV6_ADDRS_NUMOF=15 \
```
miri64: You *don't* need 15 addresses **per node**! You need at most 2 and that is already preconfigured.

MrKevinWeiss: The reasoning behind this is it will still compile/function and if a whole bunch of nodes are booked rpl can still work (even if many nodes are in range).  This means that we don't have to access each node for blacklisting.  If you would like I can say "set it to an appropriate number" so that it doesn't seem so specific.

07-multi-hop/07-multi-hop.md
MrKevinWeiss: The blacklisting seems like a good solution, especially for automation.  However, this is how I tested.  I found it was very hard to get the spacing correct and unpleasant to have to cancel and rebook an experiment because one node couldn't reach or another node could reach two.  It is not optimized but it worked.  It may also be good to test in other ways (a very controlled minimal way so we can confirm specs and a wide open way so we can see unpredictable but still valid behaviors).

miri64:
There still need to be ironed out some kinks regardless of that. Mainly, the unnecessary configuration of the number of interface IPv6 addresses `-DGNRC_NETIF_IPV6_ADDRS_NUMOF=15` is still there.

MrKevinWeiss:
@miri64 I updated to do it the l2filter_whitelist way.  This way we can keep default values for everything.  Please advise on how you would like me to proceed (should I keep this as an example and we leave the actual test more generic or make this the test?).

07-multi-hop/example_test_guide.md
```diff
@@ -0,0 +1,106 @@
+_The following is a possible way to run the tests, this is just an example_
+## Goal: Check multi-hop connectivity over IPv6
+
+### Pre-setup Steps
```
cgundogan: We also need to mention that all applications (task 1-4) need to be compiled with `USEMODULE=gnrc_pktbuf_cmd`.

07-multi-hop/example_test_guide.md
```diff
@@ -0,0 +1,113 @@
...
+## Goal: Check multi-hop connectivity over IPv6
+_We will arrange all 4 nodes in a chain topology:
+node1 <-> node2 <-> node3 <-> node4_
+
+### Pre-setup Steps
+1. Book 4 [iotlab](https://www.iot-lab.info/testbed/dashboard) M3 nodes
+
+# Task 1 and 2 Setup
+1. Open 4 instances of nodes physically close together
+`USEMODULE=gnrc_pktbuf_cmd make C tests/gnrc_udp/ BOARD=iotlab-m3
```
cgundogan: `C` => `-C`

MrKevinWeiss: ugg, thanks.

PR #88, opened 2018-11-23
10-icmpv6-error: Add ICMPv6 error message release specs

miri64:
This just adds the test descriptions for testing ICMPv6 error message behavior. The tests are based on the testing procedures I proposed for https://github.com/RIOT-OS/RIOT/pull/8594, but re-ordered to reflect the order they have in [RFC 4443](https://tools.ietf.org/html/rfc4443) and not for incremental configuration as it was in https://github.com/RIOT-OS/RIOT/pull/8594. I will provide a test script soon as a separate PR, based on #79 and the test scripts I provided in https://github.com/RIOT-OS/RIOT/pull/8594.

### DIFF
--- a/10-icmpv6-error/10-icmpv6-error.md
+++ b/10-icmpv6-error/10-icmpv6-error.md
@@ -1,0 +1,331 @@
+## Goal: Check single-hop ICMP connectivity over IPv6
+
+*Note: This task requires to send some very specificly formated IPv6 packets. We
+recommend to use [scapy] for the construction of those packets in those tests. A
+packet can be sent e.g. using (you will require root permission for sending
+using link-layer frames with `sendp()`)*
+
+```py
+sendp(Ether(dst=DST_HWADDR) / IPv6(src=SRC_IPV6, dst=DST_IPV6) / \
+      UDP(dport=DST_PORT) / PAYLOAD_BYTE_STR,
+      iface="tapbr0", timeout=0)
+```
+
+Testing procedure setup
+=======================
+All sub-tasks within this specification can be done using one `native` instance
+of the `gnrc_networking` example. For the testing process it is assumed running
+
+    $ ./dist/tools/tapsetup/tapsetup
+
+was used to create the TAP interface (including a bridge) required for the
+example. As a consequence, when talking about configuring the interface all
+Linux-side configuration is assumed to be done on the bridge `tapbr0`. The
+interface name might differ if you created it otherwise.
+
+The interface and native instance are assumed to be reset (and if necessary
+rebuild) for every task.
+
+Returned packets can be verified using a sniffing tool like Wireshark or
+[scapy].
+
+Task #01 - Destination unreachable - no route to destination
+============================================================
+### Description
+
+Solicit **Destination unreachable - no route to destination** by sending a UDP
+to a global address via a native node from a Linux host with global address.
+* Stack configuration:  IPv6 (incl. ICMPv6 error)
+* UDP Port:             48879 (no server on native node)
+* UDP Payload:          0B
+* Sender Address:       beef::1/64 (assigned to interface of Linux host,
+                        route to `beef::/64` configured on native node)
+* Destination Address:  affe::1/64 (route configured on Linux host, but not
+                        on native node)
+
+
+### Testing procedure
+
+1. Add `beef::1/64` to the TAP interface:
+
+        # ip addr add beef::1/64 dev tapbr0
+
+2. Add `affe::/64` route via the native instance on Linux host side:
+
+        # ip route add affe::/64 via "<native link local IPv6 address>" \
+                dev tapbr0
+
+3. Add `beef::/64` route via TAP interface on RIOT side:
+
+        > nib route add 6 beef::/64 "<TAP interface link-local IPv6 address>"
+
+4. Send the UDP packet as specified.
+
+### Result
+
+An ICMPv6 destination unreachable (code: 0 - no route to destination) message
+should be sent by the native node.
+
+Task #02 - Destination unreachable - Beyond scope of source address
+===================================================================
+### Description
+
+Solicit **Destination unreachable - beyond scope of source address** by sending
+a UDP to a global address via a native node from a Linux host with link-local
+address.
+* Stack configuration:  IPv6 (incl. ICMPv6 error)
+* UDP Port:             48879 (no server on native node)
+* UDP Payload:          0B
+* Sender Address:       Link-local unicast (fe80::.../64)
+* Destination Address:  affe::1 (not assigned to an interface of the native
+                        node)
+
+### Testing procedure
+
+1. Send the UDP packet as specified.
+
+### Result
+
+An ICMPv6 destination unreachable (code: 2 - beyond scope of source address)
+message should be sent by the native node.
+
+Task #03 - Destination unreachable - address unreachable (target node address)
+==============================================================================
+### Description
+
+Solicit **Destination unreachable - address unreachable** by sending a UDP to an
+address not assigned to an interface of the native node from a Linux host.
+* Stack configuration:  IPv6 (incl. ICMPv6 error)
+* UDP Port:             48879 (no server on native node)
+* UDP Payload:          0B
+* Sender Address:       Link local unicast (fe80::.../64)
+* Destination Address:  fe80::1 (not assigned to an interface of the native
+                        node)
+
+### Testing procedure
+
+1. Send the UDP packet as specified.
+
+### Result
+
+An ICMPv6 destination unreachable (code: 3 - address unreachable) message should
+be sent by the native node.
+
+Task #04 - Destination unreachable - address unreachable (neighbor cache miss)
+==============================================================================
+### Description
+
+Solicit **Destination unreachable - address unreachable** by sending a UDP to a
+misconfigured route on the native node from a Linux host.
+* Stack configuration:  IPv6 (incl. ICMPv6 error)
+* UDP Port:             48879 (no server on native node)
+* UDP Payload:          0B
+* Sender Address:       beef::1/64 (assigned to interface of Linux host,
+                        route to `beef::/64` configured on native node)
+* Destination Address:  affe::1/64 (route configured on Linux host and on native
+                        node, the latter to a non-existing neighbor)
+
+### Testing procedure
+
+1. Add `beef::1/64` to the TAP interface:
+
+        # ip addr add beef::1/64 dev tapbr0
+
+2. Add `affe::/64` route via the native instance on Linux host side:
+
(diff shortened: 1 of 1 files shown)


10-icmpv6-error/10-icmpv6-error.md
```diff
@@ -0,0 +1,176 @@
+## Goal: Check single-hop ICMP connectivity over IPv6
+
+*Note: This task requires to send some very specificly formated IPv6 packets. We
+recommend to use [scapy] for the construction of those packets in those tests.*
+
+Task #01 - Destination unreachable - no route to destination
```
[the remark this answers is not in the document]
miri64: I intentionally wanted to keep these specs implementation independent and then provide the actual code in a testing script (as I hinted at in OP).

miri64:
I added another test-case which isn't really ICMPv6 error triggering but definitely goes into the direction "weirdly formatted packets send to the stack" which can be implemented with `scapy` and native as well:

```py
ip = IPv6(src=src_ipv6, dst=dst_ipv6)
i = 0
while (((i + 1) * 40) + 8) < 1500:
    ips = ip
    for _ in range(i):
        ips /= ip
    sendp(Ether(dst=ether_dst) / ips / ICMPv6EchoRequest(seq=i), iface="tapbr0")
    i += 1
    print(i - 1)
    time.sleep(0.001)
```

[Here you can find a fully executable version of that script](https://gist.github.com/miri64/a79732b587a69cb89b6afc0b4911f713)

miri64:
@aabadie ping?

10-icmpv6-error/10-icmpv6-error.md
```diff
@@ -0,0 +1,188 @@
...
+============================================================
+### Description
+
+Solicit **Destination unreachable - no route to destination** by sending a UDP
```
aabadie: It is a bit misleading to have mentions to UDP in an ICMPv6 test. I think it should rather be `ICMPv6EchoRequest` instead of `UDP` here and in other places, no?

aabadie:
I think it makes sense to add these test cases to the release specs, but I also think that, as is, it is difficult to understand how setup everything to correctly test this.
After reading the testing procedure in https://github.com/RIOT-OS/RIOT/pull/8594, I think I'm starting to understand things. Why not having everything in the same place ?
I always thought that those release specs description were too high level and that it would be great to provide some associated setup/commands. Otherwise, only a very few people know how to run them (think of the bus factor here).

aabadie:
Maybe adding a `Procedure` section between `Description` and `Result` would improve the situation ?

10-icmpv6-error/10-icmpv6-error.md
miri64: a) I don't really get what is misleading about this? These are tests to trigger ICMPv6 error messages. What kind of packet triggers them should be irrelevant (and in fact IMHO it should (but must not) at least for the tests be disjunct from ICMP so using e.g. UDP helps with that).
b) From a practical standpoint ICMPv6EchoRequest has the disadvantage that we can't easily register to it (it would require a dedicated command i.e. test application). `udp server` is up and ready in `gnrc_networking`. Just listening for an ICMPv6EchoReply could also be viable, but it gives no guarantee on which node it was handled (from a pure testing perspective of course were no assumptions can be made, in practice it should of course be the node targeted).

miri64:
> After reading the testing procedure in [RIOT-OS/RIOT#8594](https://github.com/RIOT-OS/RIOT/pull/8594), I think I'm starting to understand things. Why not having everything in the same place ?

For a third time (see https://github.com/RIOT-OS/Release-Specs/pull/88#discussion_r235990818):

> I intentionally wanted to keep these specs implementation independent and then provide the actual code in a testing script (as I hinted at in OP).

aabadie:
> I intentionally wanted to keep these specs implementation independent and then provide the actual code in a testing script (as I hinted at in OP).

That doesn't answer what I'm looking for. As is, if I just read this README, I have no idea how to test it.

miri64:
> I always thought that those release specs description were too high level and that it would be great to provide some associated setup/commands.

IMHO the point of a *specification* is to be high level. Everything else goes already more into the direction of test implementation.

miri64:
> > I intentionally wanted to keep these specs implementation independent and then provide the actual code in a testing script (as I hinted at in OP).
> 
> That doesn't answer what I'm looking for. As is, if I just read this README, I have no idea how to test it.

Then please provide some directions how I might help with that without restricting these specs to a specific implementation.

aabadie:
> Then please provide some directions how I might help with that without restricting these specs to a specific implementation.

See https://github.com/RIOT-OS/Release-Specs/pull/88#issuecomment-454683207 above. Might be a good starting point. We already provide this in PRs why not backport the testing procedure to release specs ?

10-icmpv6-error/10-icmpv6-error.md
miri64: Also: ICMPv6 has no ports, so testing Task 5 without UDP would be quite impossible ;-).

miri64:
> > Then please provide some directions how I might help with that without restricting these specs to a specific implementation.
> 
> See [#88 (comment)](https://github.com/RIOT-OS/Release-Specs/pull/88#issuecomment-454683207) above. Might be a good starting point. We already provide this in PRs why not backport the testing procedure to release specs ?

So something like (for task 1)

```markdown
### Procedure
1. Compile `examples/gnrc_networking` for `native`
2. Create a TAP interface (if not already done).
   E.g. `./dist/tools/tapsetup/tapsetup` (creates a TAP bridge as well)
3. Execute `examples/gnrc_networking`
4. Assign `beef::1/64` to the TAP bridge / TAP interface.
   E.g. `ip addr add beef::1/64 dev tapbr0`
5. Configure a route to `affe::/64` via the TAP bridge / TAP interface to the
   `example/gnrc_networking` instance.
   E.g. `ip route add affe::/64 via <link-local address of native instance> dev tapbr0`
6. Use your favorite method for sending UDP packets from the Linux host to
   `[affe::1]:48879` with an empty payload.
```

aabadie:
> So something like (for task 1)

Not sure this should be copied in all sub-task but it could be in a common setup section at the beginning of the file and an internal link added to each sub-task.

I also like what's done in #77 where the task scripts are also provided with some notes on how to play the task.

Of course it would be better to automatize all that (using pytest, which is my highly preferred solution compared to others proposed) but that doesn't prevent to provide a manual procedure.

miri64:
> Of course it would be better to automatize all that (using pytest, which is my highly preferred solution compared to others proposed) but that doesn't prevent to provide a manual procedure.

I think the discussion regarding framework setup is still ongoing in https://github.com/RIOT-OS/RIOT/issues/10241 (yet another reason why I didn't want to provide an implementation yet) and I know you have a strong opinion against RFW (as I did [but that changed](https://github.com/RIOT-OS/RIOT/issues/10241#issuecomment-441254895)). I think your comments there regarding your test framework of choice would be very welcome there.

miri64:
> Of course it would be better to automatize all that (using pytest, which is my highly preferred solution compared to others proposed) but that doesn't prevent to provide a manual procedure.

Again, these are specifications. That means apart from giving us (the contributors) an agreed upon way to test the release they also provide a user with the information "This is what has been tested for the release". I'm a bit worried extensive testing descriptions (especially if implementation specific) might confuse more in the latter case than be helpful.

miri64:
(+ lead to testers doing the same predescribed thing over and over again without even thinking about it.... in the past we found bugs during the release testing *because* testers did or tried something different then other testers before them).

aabadie:
> in the past we found bugs during the release testing because testers did or tried something different than other testers before them

Providing or suggesting a testing procedure would increase the number of potential testers. Like it is now, with almost no guide and specifications that are becoming more and more vague/complex to setup and test, it ends up in this to be only testable by the person who initially wrote the specification (or maybe the same 2 or 3 people with enough knowledge), because he/she knows how to do it. This breaks your argument since one will very quickly fall into "things being tried always the same way".
#85 is also going in the direction of helping testers starting with those non straight-forward (for uninitiated) specifications.

miri64:
Done

10-icmpv6-error/10-icmpv6-error.md
```diff
@@ -0,0 +1,323 @@
...
+was used to create the TAP interface (including a bridge) required for the
+example. As a consequence, when talking about configuring the interface all
+Linux-side configuration is assumed to be done on the bridge `tapbr0`. The
+interface name might differ if you created it otherwise.
+
+The interface and native instance are assumed to be reset (and if necessary
+rebuild) for every task.
+
+Returned packets can be verified using a sniffing tool like Wireshark or
+[scapy].
```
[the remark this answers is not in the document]
miri64: This is a reference. `[scapy]` is defined below and the link is generated. Have a look at the rendered output ;-). This way you can read the text raw better.

10-icmpv6-error/10-icmpv6-error.md
```diff
@@ -0,0 +1,330 @@
+## Goal: Check single-hop ICMP connectivity over IPv6
+
+*Note: This task requires to send some very specificly formated IPv6 packets. We
+recommend to use [scapy] for the construction of those packets in those tests. A
+packet can be sent e.g. using (you will require root permission for sending
+using link-layer frames with `sendp()`)*
+
+```py
+sendp(Ether(dst=DST_HWADDR) / IPv6(src=SRC_IPV6, dst=DST_IPV6) / UDP(dport=DST_PORT),
```
aabadie: This is much better. Stupid (or lazy guy) question: how to set a payload of 1452 as required [here](https://github.com/RIOT-OS/Release-Specs/pull/88/files#diff-cce1e7e335145eaef7f0b065d1b728bdR182) ?

Would the following work ?
```py
sendp(Ether(dst=DST_HWADDR) / IPv6(src=SRC_IPV6, dst=DST_IPV6) / UDP(dport=DST_PORT) / "x" * 1452 ,iface="tapbr0", timeout=0)
```

miri64: Yepp

miri64: Added something like that to the doc ;-).

miri64:
Since I gave this the number 10, but #77 is not merged yet, how do we handle this? Should I renumber, should we wait for #77, or just merge and accept that there will be a hole for a while?

aabadie:
I'd like to also merge 09-coap for this release but I want to test it before.

miri64:
Please also consider https://github.com/RIOT-OS/Release-Specs/pull/90 for the release. It took me quite some elbow grease to get those bugs fixed and I'd hate a regression to be introduced ;-).

PR #91, opened 2019-01-24
task06: add specification for neighbor address resolution errors

miri64:
Adds a regression test for the bug fixed in https://github.com/RIOT-OS/RIOT/pull/10860 (and as an extra also the same kind of test for 6Lo-based nodes).

### DIFF
--- a/06-single-hop-udp/06-single-hop-udp.md
+++ b/06-single-hop-udp/06-single-hop-udp.md
@@ -47,7 +47,7 @@ Sending UDP from one native node to a non-existent neighbor.
 * Interval:               0us
 * Port:                   1337
 * Payload:                8B
-* Destination Address:    Non-existent Link local unicast (e.g fe80::bd:b7ec)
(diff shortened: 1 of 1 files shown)


miri64:
This wasn't squashed...

aabadie:
> This wasn't squashed...

Ah, sorry, I didn't check the commit history.

aabadie:
Maybe squash and force push ? There is high probability that this won't affect anyone.

miri64:
> Maybe squash and force push ? There is high probability that this won't affect anyone.

While I agree, that it probably won't hurt anyone there is still the likelihood it will hurt someone, so I'm against force pushing into a public master ;-).

miri64:
We should however learn and establish some squash check here (and in riotdocker) as well ;-)

aabadie:
> I'm against force pushing into a public master

so am I :)

> We should however learn and establish some squash check here (and in riotdocker) as well ;-)

Maybe TravisCI could help

PR #96, opened 2019-01-29
task10: add requirement for MTU field of Packet Too Big message

miri64:
Since the MTU field of a Packet Too Big message is quite important for Path MTU Discovery (see [RFC 8201](https://tools.ietf.org/html/rfc8201)) we should check its value in our release tests.


miri64:
From [section 3](https://tools.ietf.org/html/rfc8201#section-3):

> Upon receipt of such a [Packet Too Big] message, the
>  source node reduces its assumed PMTU for the path based on the MTU of
>  the constricting hop as reported in the Packet Too Big message.

PR #97, opened 2019-02-04
task08.3: add testing procedures

miri64:
This provides the testing procedures for task 8.3. It also adapts the specifications to the tests that actually happened in the past (UDP exchange, not ICMPv6 echo request/reply as the latter are much harder to realize with contiki)

### DIFF
--- a/08-interop/08-interop.md
+++ b/08-interop/08-interop.md
@@ -20,6 +20,96 @@ Task #03 - UDP exchange between iotlab-m3 and Contiki
 
 UDP packet exchange between an iotlab-m3 node running RIOT and a Contiki node.
 
+### Testing procedure
+1. Get your hands on a [CC2538DK evaluation board][cc2538dk]
+1. Clone the Contiki repository to your machine
+   ```sh
+   git clone http://github.com/contiki-os/contiki.git && cd contiki/
+   ```
+2. Go to the `udp-ipv6-echo-server` for the CC2538DK:
+   ```sh
+   cd examples/cc2538dk/udp-ipv6-echo-server
+   ```
+3. Apply the following patch
+   ```sh
+   echo 'diff --git a/examples/cc2538dk/udp-ipv6-echo-server/Makefile b/examples/cc2538dk/udp-ipv6-echo-server/Makefile
+   index 5bbbdd6..92ec1c5 100644
+   --- a/examples/cc2538dk/udp-ipv6-echo-server/Makefile
+   +++ b/examples/cc2538dk/udp-ipv6-echo-server/Makefile
+   @@ -1,8 +1,12 @@
+   +DEFINES+=PROJECT_CONF_H=\"project-conf.h\"
+    CONTIKI_PROJECT = udp-echo-server
+
+    all: $(CONTIKI_PROJECT)
+
+    CONTIKI = ../../..
+   +CONTIKI_WITH_RIME = 0
+    CONTIKI_WITH_IPV6 = 1
+    CFLAGS += -DUIP_CONF_ND6_SEND_NS=1
+   +CFLAGS += -DRF_CHANNEL=26
+   +CFLAGS += -DIEEE802154_CONF_PANID=0x23
+    include $(CONTIKI)/Makefile.include
+   diff --git a/examples/cc2538dk/udp-ipv6-echo-server/project-conf.h b/examples/cc2538dk/udp-ipv6-echo-server/project-conf.h
+   new file mode 100644
+   index 0000000..1718ded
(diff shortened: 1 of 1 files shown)


08-interop/08-interop.md
```diff
@@ -14,12 +14,98 @@ Task #02 - ICMPv6 echo between iotlab-m3 and Linux with 6LowPAN
...
+1. Clone the Contiki repository to your machine
+   ```sh
+   git clone http://github.com/contiki-os/contiki.git && cd contiki/
+   ```
+2. Go to the `udp-ipv6-echo-server` for the CC2538DK:
+   ```sh
+   cd examples/cc2538dk/udp-ipv6-echo-server
+   ```
+3. Apply the following patch
+   ```sh
```
kYc0o: It shouldn't be ```diff here? I think it's better formatted that way, unless you wanted to paste the actual comand to apply the patch.

miri64: Yes, I wanted to paste the command to apply the patch. This way the tester just need to copy&paste ;-).

miri64: However, I added the missing pipe below ;-D

kYc0o:
Well I would ack but I suppose it needs to be tested and I don't have the hardware, I think someone at HAW Hamburg has it, maybe @smlng ?

miri64:
Fixed some nits I found myself while running the tests for RC3.

miri64:
Squashed. And ping? I think it is good if we merge this before 2019.04-RC1 is out ;-)

danpetry:
> And ping?

I thought the point was to remove this? Please clarify

miri64:
> I thought the point was to remove this? Please clarify

Remove what? The PR? Where did you read that?

08-interop/08-interop.md
danpetry: I used this to copy and paste...

08-interop/08-interop.md
```diff
@@ -14,12 +14,101 @@ Task #02 - ICMPv6 echo between iotlab-m3 and Linux with 6LowPAN
...
+
+   You might have seen this address also during flashing
+
+        Primary IEEE Address: 00:12:34:56:78:9A:BC:DE
+
+   You can get the IPv6 link local address of the node by toggling the second
+   bit of the MSB of the address and prepending `fe80::`. So for the example
+   above that would be `fe80::0212:3456:789a:bcde`.
+7. For the `iotlab-m3` side just use [gnrc_networking] or the [gnrc_udp] test
+   application, whichever you prefer. There start a UDP server on port 3000 with
```
danpetry: Flashing the iotlab-m3, I needed to specify PORT=/dev/ttyUSB3. 

The flashing port and term port are different from each other on the M3, and assigned to 0 and 1 by default. So if you've used USB0 and 1 for the CC2538DK, and still have it connected, you need to specify manually. Furthermore there are two M3 ports, so I had to try USB2 before USB3 worked. Not quite sure why, because I thought I would have had to specify both the flashing and term ports, but it did.

You can also obv unplug the CC2538DK and then not specify the port with the M3, then plug the CC2538DK back in, but that's not very automatable...

danpetry:
> Remove what? The PR? Where did you read that?

No, to remove the ping, i.e the ICMPv6 echo. You said "and ping". Never mind...

miri64:
> No, to remove the ping, i.e the ICMPv6 echo. You said "and ping". Never mind...

That was not related to ICMPv6 echo, but a notch to @kYc0o and @smlng to review this ;-)...

08-interop/08-interop.md
miri64: > Flashing the iotlab-m3, I needed to specify PORT=/dev/ttyUSB3.

This can't be specified, as this is highly dependent on the local setup of the tester. I'd say: let's assume the tester knows about this.



> The flashing port and term port are different from each other on the M3, and assigned to 0 and 1 by default. So if you've used USB0 and 1 for the CC2538DK, and still have it connected, you need to specify manually. Furthermore there are two M3 ports, so I had to try USB2 before USB3 worked. Not quite sure why, because I thought I would have had to specify both the flashing and term ports, but it did.
> 
> You can also obv unplug the CC2538DK and then not specify the port with the M3, then plug the CC2538DK back in, but that's not very automatable...

Also more a problem with `iotlab-m3` (or FTDI-based flashers?) in general that can be assumed a tester knows, not important for this test.

danpetry:
Yes, I get that. Next time I'll make sure my joke submission is higher quality before presenting it.

PR #99, opened 2019-02-04
03-single-hop-ipv6-icmp: fix typo in task 4

aabadie:


aabadie:
Let's auto-ACK and merge.

PR #103, opened 2019-02-10
11-lorawan: add release specs for lorawan

aabadie:
Add release specs for the LoRaWAN features. It would be great to have them for testing the 2019.01 RC3.

### DIFF
--- a/11-lorawan/11-lorawan.md
+++ b/11-lorawan/11-lorawan.md
@@ -1,0 +1,194 @@
+## Goal: Verify LoRaWAN networking
+
+Task #01 - LoRaWAN example
+==========================
+
+### Description
+
+LoRaWAN example is fully functional
+
+### Testing procedure
+
+Build and flash the `examples/lorawan` application with the following command
+
+    $ make BOARD=<board> DEVEUI=<device eui> APPEUI=<application eui> APPKEY=<application key> -C examples/lorawan flash term
+
+_Note:_ The application can be tested on IoT-LAB with a device register on
+TheThingsNetwork backend: start an experiment with one of the st-lrwan1 boards
+and pass `IOTLAB_NODE=auto-ssh` to the make command.
+
+### Result
+
+With valid device EUI, application EUI and application key, the example
+application can join the backend. Messages are received on the backend every
+~20s.
+
+
+Task #02 - OTAA join procedure
+==============================
+
+### Description
+
+OTAA join procedure and uplink with datarates indexes 0, 3, 5 (others are optional)
+
+### Testing procedure
+
+- Build and flash the `tests/pkg_semtech-loramac`
+
+      $ make BOARD=<board> -C tests/pkg_semtech-loramac flash term
+
+- Configure the devive EUI, application EUI, application key
+
+      > loramac set deveui 0011223344556677
+      > loramac set appeui 0011223344556677
+      > loramac set appkey 00112233445566770011223344556677
+
+- Configure the desired datarate index
+
+      > loramac set dr <datarate index>
+
+- Join the network using OTAA
+
+      > loramac join otaa
+
+- Send confirmed messages on port 123
+
+      > loramac tx <message> cnf 123
+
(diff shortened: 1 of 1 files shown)


11-lorawan/11-lorawan.md
```diff
@@ -0,0 +1,188 @@
...
+
+- Send confirmed messages on port 123
+
+      > loramac tx <message> cnf 123
+
+- Send unconfirmed messages on port 42
+
+      > loramac tx <message> uncnf 42
+
+- Reboot the node after each datarate test
```
jia200x: I would add here that's required to reset the frame counters in TTN. Otherwise the backend will ignore all uplink packets after reset

jia200x:
besides my tiny comment, this test makes sense to me. Great work!

11-lorawan/11-lorawan.md
aabadie: I wrote this with STM32 L0/L1 in mind: they provide an EEPROM so the uplink counter is stored automatically and there's no need to reset the frame counter in TTN. I'll add a note about that.

aabadie:
@jia200x, I added a note about frame counters. Directly squashed.

11-lorawan/11-lorawan.md
```diff
@@ -0,0 +1,194 @@
...
+
+- Build and flash the `tests/pkg_semtech-loramac`
+
+      $ make BOARD=<board> -C tests/pkg_semtech-loramac flash term
+
+- Configure the devive EUI, application EUI, application key
+
+      > loramac set deveui 0011223344556677
+      > loramac set appeui 0011223344556677
+      > loramac set devkey 00112233445566770011223344556677
```
jia200x: s/devkey/appkey

11-lorawan/11-lorawan.md
```diff
@@ -0,0 +1,194 @@
...
+      > loramac tx <message> uncnf 42
+
+- Reboot the node after each datarate test
+
+### Result
+
+All joins are successful, messages are sent with success and correctly received
+on the LoRaWAN backend (same datarate, same port).
+TX attempts within the dutycycle are skipped (very likely to happen with low
+datarates). Messages sent from the backend to the device are received.
```
jia200x: with higher DR (means lower spreading factor) the time on air is smaller. So, I guess it should be the opposite:

```
...(very likely to happen with high datarates)
```

11-lorawan/11-lorawan.md
```diff
@@ -0,0 +1,194 @@
...
+      > loramac tx <message> uncnf 42
+
+- Reboot the node after each datarate test. _Note:_ When using a board without
+  EEPROM support, the frame counters must also be reset on the LoRaWAN backend.
+
+### Result
+
+All messages are sent with success and correctly received on the LoRaWAN
+backend (same datarate, same port).
+TX attempts within the dutycycle are skipped (very likely to happen with low
```
jia200x: same as above

jia200x:
(sorry, didn't spot those errors before)

11-lorawan/11-lorawan.md
aabadie: Ok, this is confusing, I meant low datarate indexes (DR0 for instance)

aabadie: Well, now I'm re-reading it, I think it's fine as it now. Low datarates, DR0, will have a longer duty-cycle, so one has to wait more between 2 TX, that's why TX attempts have higher chances to be skipped with lower DR.

jia200x: ah! you are right. I read it again and of course makes sense

11-lorawan/11-lorawan.md
aabadie: Fixed (with another one below)

jia200x:
ACK

jia200x:
thanks!

aabadie:
Thanks @jia200x !

PR #106, opened 2019-02-13
task03: fix typo in task 5

miri64:
Fixes a small but significant typo in Task 3 ;-)


aabadie:
Already done (who is release manager btw ??!!!)

miri64:
You are, but I (co-)introduced this bug, so I feel responsible ;-).

PR #109, opened 2019-02-14
task07: remove redundant ping

miri64:
This is more of an RFC. I think the second ping command is redundant, since the ping exchange is already bidirectional (so compared to UDP we don't have to check "the other direction"). However, I see the benefit in having a redundant ping from the other end to mitigate errors in the setup.

### DIFF
--- a/07-multi-hop/example_test_guide.md
+++ b/07-multi-hop/example_test_guide.md
@@ -31,7 +31,6 @@ with static routes.
 
 1. Follow Task 1 and 2 setup
 2. Ping node 4 from node 1, Node 1:`ping6 -c 100 -s 50 -i 100 abcd::2`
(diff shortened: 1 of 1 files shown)


MrKevinWeiss:
I think it doesn't hurt leaving it in...

danpetry:
Well, I think it's fairly minor but I'm personally for making things as efficient as possible (not removing tests obv!) in the release tests area. So I'd say remove, myself.

>  to mitigate errors in the setup

I'm not sure what errors these could be, could you please explain (on or offline)?

miri64:
Rebased to current master. You wanna merge or can I hit the friendly green button?

MrKevinWeiss:
Now nobody will know that I don't understand ping :smile:

PR #119, opened 2019-04-08
Include instruction to run tapsetup

danpetry:
Tell the tester explicitly to set up the tap interfaces if they are running tests for native.


miri64:
Duplicate of #125 and #130

Issue #113, opened 2019-04-05
Release 2019.04 - RC1


danpetry:
This issue is for discussion related to testing and bugfixing of [Release Candidate 1 of the 2019.04](https://github.com/RIOT-OS/RIOT/releases/tag/2019.04-RC1) release.

To track testing, please use the tracking spreadsheet, which contains the list of tests and a column for your initials if you're testing something, and pass/fail status for your test. This should hopefully be a better way of capturing information than a checklist and inline conversation, let me know if it's not.

https://docs.google.com/spreadsheets/d/18k5rijHnoFEC3AZA5Ur3_nRjwban-5na6R2bkDu2Jac/edit?usp=sharing

There is a "test failures and bugfixes" tab in the spreadsheet, which is there to capture the progress of bugfixes coming out of the tests. This is optional for you, as I'll keep it updated, but hopefully will be also useful.

There are also folders for you to drop test artefacts in, if there are any:

https://drive.google.com/drive/folders/14EwQPQM7zN0Go5SJHBtIOBF6qN8L7vig?usp=sharing

miri64:
Why is 1.1 marked as failed? As far as I can see the linked Murdock output only fails on the tests for `tests/pkg_c25519` but not on the compiling of any of those applications.

danpetry:
It says "failures fixed" next to it, Koen submitted a PR for the fix which got backported so this indicates that. Maybe it's not clear enough, I'll change

danpetry:
Ok thank you for the clarification!

jia200x:
Test 11.3 (LoRaWAN abp) fails. It freezes:
```
main(): This is RIOT! (Version: 2019.07-devel-HEAD)
All up, running the shell now
> loramac set nwkskey B74B805FAFC3E10B81E7A9015E67B43C
loramac set nwkskey B74B805FAFC3E10B81E7A9015E67B43C
> loramac set appskey 59A6C0AA0870E53CCAD8E4EB7647D10C
loramac set appskey 59A6C0AA0870E53CCAD8E4EB7647D10C
> loramac set rx2_dr 3
loramac set rx2_dr 3
> loramac join abp
loramac join abp
Join procedure succeeded!
> loramac tx hola
loramac tx hola
help

/* Crickets... */
```

miri64:
[...]

PR #121, opened 2019-04-10
task04: change timings for large payload test

miri64:
Due to fragmentation the average round trip time of a link-local ping
is about 140ms (depending on distance). This puts a strain on both
packet buffer and reassembly buffer on the sending nodes as it has to
handle at least two packets while it is also working on a reply.

Add to that that the `at86rf2xx` driver is only able to send when it is
not in receive state (and blocks until it leaves) the pinged node has
problems to get out the data it wants to send the echo replies while it
is being spammed with fragments containing follow-up echo requests.

Because the test is mainly about testing fragmentation, not
fragmentation under stress, I suggest we tweak the test parameters a
bit so the test takes the same time, but with less stress on the node.

Also see https://github.com/RIOT-OS/Release-Specs/issues/113#issuecomment-481733314

### DIFF
--- a/04-single-hop-6lowpan-icmp/04-single-hop-6lowpan-icmp.md
+++ b/04-single-hop-6lowpan-icmp/04-single-hop-6lowpan-icmp.md
@@ -40,7 +40,7 @@ ICMPv6 echo request/reply exchange between two nodes.
 * Stack configuration:    6LoWPAN (default)
 * Channel:                26
 * Count:                  500
-* Interval:               200ms
+* Interval:               300ms
 * Payload:                1kB
 * Destination Address:    Link local unicast (fe80::.../64)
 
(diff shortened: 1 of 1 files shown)


danpetry:
With the automated script (i.e. using IoT-LAB), this works if you put the delay up to 300ms. But not 200 ms. I guess there's a longer round trip. Suggest this time instead?

miri64:
> With the automated script (i.e. using IoT-LAB), this works if you put the delay up to 300ms. But not 200 ms. I guess there's a longer round trip. Suggest this time instead?

Mh... Would have been good to provide some output with that, so I could assess. But I guess 300ms should be fine. Will fix

miri64:
Done

danpetry:
By my reckoning, for the same reason, the timing should be increased also on:

- test 05 task 02 (static routing test single hop, iotlab-m3, default route)
- test 05 task 04 (static routing test single hop, iotlab-m3, static route)

Re the iot-lab scripts, I'm having a look at it. But 300ms interval works ok for on-desk tests.

miri64:
> By my reckoning, for the same reason, the timing should be increased also on:
> 
> * test 05 task 02 (static routing test single hop, iotlab-m3, default route)
> * test 05 task 04 (static routing test single hop, iotlab-m3, static route)

Let's do that in a separate PR

danpetry:
> Let's do that in a separate PR

Will open in a minute

PR #127, opened 2019-07-01
02-tests: replace with RIOT compile_and_test_for_board.py

cladmi:
The script is now available in `RIOT` so replace mention in the
documentation. Also update 99-compile-and-test-one-board.

You can check the path in the documentation is indeed valid.

```
ls RIOT/dist/tools/compile_and_test_for_board/README.md
RIOT/dist/tools/compile_and_test_for_board/README.md
ls RIOT/dist/tools/compile_and_test_for_board/compile_and_test_for_board.py
RIOT/dist/tools/compile_and_test_for_board/compile_and_test_for_board.py
```


cladmi:
Thanks

PR #125, opened 2019-04-25
02-tests: Specify that tapsetup should be run first

danpetry:
This prevents the test engineer from running the tests and then
realising that the test requires the tap bridge to be up.

PR #130, opened 2019-07-15
02-tests: Specify that tapsetup should be run first

miri64:
This prevents the test engineer from running the tests and then
realising that the test requires the tap bridge to be up.

Reopen of #125.


cladmi:
This reverted the `compile_and_test_for_board.py` script path to the old one before c93126f

miri64:
Oops, fixed in #131.

PR #131, opened 2019-07-15
task02: Fix rebase error of #130

miri64:
Fixes rebase error from #130.


cladmi:
I was doing the script handling at the same time :D 
The path is now correct again.

It also precises that it is only for `native`.

MrKevinWeiss:
unless the regular old contributor @cladmi has something to say :smile:

Issue #137, opened 2019-10-04
RFC: Add tests for IPv6 fragmentation


miri64:
With IPv6 fragmentation and reassembly merged (https://github.com/RIOT-OS/RIOT/pull/11623 and https://github.com/RIOT-OS/RIOT/pull/11596 respectively) during the 2019.10 release cycle, I think it makes sense to extend the release specifications for packets of sizes (MTU, 2*MTU]. Specifically I would extend `03`, `04` and `07` with one task each to ping with packets with e.g. 2 KiB payload size.

All weird corner cases should already be covered by the `tests/gnrc_ipv6_ext_frag` application (which _should_ be covered by `02`, but who knows, since it requires `sudo` and people tend to skip tests that are not auto-run ;-P). However, I prefer to have at least some "in-the-wild" tests done during the release testing.

TBH, I personally never tested IPv6 + 6LoWPAN fragmentation (but AFAIK @benpicco did during the testing of https://github.com/RIOT-OS/RIOT/pull/11623 *[edit]after rereading his review comment I'm less sure[/edit]*), so we have to see if the `04` extension works as expected and I don't believe multihop (`07`) was properly tested yet (`tests/gnrc_ipv6_ext_frag` tests forwarding of fragments though). I would thus suggest that the `04` and `07` extensions should be marked as experimental at least for the first RC they are tested in, in case some weird side effects arise.

leandrolanzieri:
+1 for extending the tests and making those tasks experimental for now in `04` and `07`, I think we should test this feature in multiple conditions.

On a separate matter, tasks 4.(5-8) still remain experimental? They seem to have been passing, @aabadie ?

miri64:
Ok, then I prepare a PR with the specifications.

miri64:
Good that we test this. Pinging with 2KiB already maxes out the packet buffer xD

miri64:
> On a separate matter, tasks 4.(5-8) still remain experimental? They seem to have been passing, @aabadie ?

Let's discuss this in a separate issue, please.

kb2ma:
I just merged #137. With @leandrolanzieri  and @fjmolinas agreement here, let's close this issue.

PR #138, opened 2019-10-07
task0{3,4,7}: Add tests for IPv6 fragmentation

miri64:
See https://github.com/RIOT-OS/Release-Specs/issues/137

Requires https://github.com/RIOT-OS/RIOT/pull/12375 for simple deployment


kb2ma:
Overall, this addition looks appropriate to me. I think the multi-hop UDP test is a valuable real world scenario.

kb2ma:
I ran the new tests for tasks 3 and 4. My only question was which example/test to run, and I see the example guide already specifies tests/gnrc_udp. In the interest of time I did not run task 7, but the instructions look clear enough. The new tests seem like a reasonable extension of the existing tests.

kb2ma:
My only concern is generically setting the larger pktbuf size in test/gnrc_udp, as commented in RIOT-OS/RIOT#12375. Let's resolve that, and then we can resolve here.

PR #149, opened 2020-03-19
08-Interop: Add task to exchange UDP packets with a Zephyr node

leandrolanzieri:
This adds a new task to the interop specifications, where UDP packets should be exchanged between a RIOT node and a Zephyr node. The task is similar to what we have with Contiki in task 03. As setup is not so hard and the example works out of the box I though it may be good to add this OS to our testings.

I included detailed instructions on how to setup everything and get it to run.

### DIFF
--- a/08-interop/08-interop.md
+++ b/08-interop/08-interop.md
@@ -169,3 +169,123 @@ Task #10 - UDP between lwIP and emb6 on iotlab-m3
 Link-local UDP over IPv6 packet exchange (payload length 8) between an iotlab-m3
 node running RIOT with lwIP and an iotlab-m3 node running RIOT with emb6 (in
 both directions).
+
+Task #11 - UDP exchange between iotlab-m3 and Zephyr
+=====================================================
+### Description
+
+UDP packet exchange between an iotlab-m3 node running RIOT and a samr21-xpro
+node running Zephyr.
+
+### Testing procedure
+
+#### Setting up the Zephyr node
+
+1. Follow the
+['Getting Started'](https://docs.zephyrproject.org/latest/getting_started/index.html)
+guide on Zephyr's documentation.
+2. You will be using the `echo_server` example application. To compile and
+   flash, connect the samr21-xpro board and in zephyr's root directory run:
+
+   ```sh
+   west build -p auto -b atsamr21_xpro samples/net/sockets/echo_server -- -DOVERLAY_CONFIG=overlay-802154.conf
+   west flash
+   ```
+3. If everything is OK you should be able to connect to the serial output use
+   any terminal program. For example:
+   ```sh
+   minicom -D /dev/ttyACM0 -o
+   ```
+4. By pressing tab you should be able to see all available commands. Get the
+   node's IPv6 by running the following on its shell:
+
+   ```sh
+   # On samr21-xpro:zephyr node
+   uart:~$ net ipv6
+
+   IPv6 support                              : enabled
+   IPv6 fragmentation support                : disabled
+   Multicast Listener Discovery support      : enabled
+   Neighbor cache support                    : enabled
+   Neighbor discovery support                : enabled
+   Duplicate address detection (DAD) support : enabled
+   Router advertisement RDNSS option support : enabled
+   6lo header compression support            : enabled
+   Max number of IPv6 network interfaces in the system          : 1
+   Max number of unicast IPv6 addresses per network interface   : 3
+   Max number of multicast IPv6 addresses per network interface : 4
+   Max number of IPv6 prefixes per network interface            : 2
+
+   IPv6 addresses for interface 0x20007d60 (IEEE 802.15.4)
+   =======================================================
+   Type            State           Lifetime (sec)  Address
+   autoconf        preferred       infinite        fe80::d419:100:7ae4:9b3b/128
+   manual          preferred       infinite        2001:db8::1/128
+   ```
+
+   Similarly, to get the IEEE 802.15.4 PAN ID and channel, run:
+
+   ```sh
+   # On samr21-xpro:zephyr node
+   uart:~$ ieee802154 get_pan_id
+   PAN ID 43981 (0xabcd)
+
+   uart:~$ ieee802154 get_chan
+   Channel 26
+   ```
+
+#### Setting up the RIOT node
+
+1. Flash the [gnrc_networking] example or the [gnrc_udp] test to the iotlab-m3 board
+   test.
+2. Set the channel and PAN ID to the same values as the Zephyr node:
+   ```sh
+   # On iotlab-m3:riot node
+   > ifconfig 6 set chan 26
+   ifconfig 6 set chan 26
+   success: set channel on interface 6 to 26
+
+   > ifconfig 6 set pan_id 0xabcd
+   ifconfig 6 set pan_id 0xabcd
+   success: set network identifier on interface 6 to 0xabcd
+   ```
+3. The UDP echo server application on the Zephyr node will be listening on port
+   4242, and echoing to the source port of the incoming packet. As both
+   [gnrc_networking] example and [gnrc_udp] test applications use the
+   destination port as the source port for the `udp send` command, we need to
+   start a server on port 4242:
+
+   ```sh
+   # On iotlab-m3:riot node
+   > udp server start 4242
+   udp server start 4242
+   Success: started UDP server on port 4242
+   ```
+4. Send packets to the echo server in the Zephyr node, they should be echoed and
+   printed on the RIOT side. Zephyr node will inform of the received packet as
+   well on the shell.
+
+   ```sh
+   # On iotlab-m3:riot node
+   > udp send fe80::d419:100:7ae4:9b3b 4242 "RIOT Testing!"
+   udp send fe80::9c1a:100:42e5:9b3b 4242 "RIOT Testing!"
+   Success: sent 13 byte(s) to [fe80::9c1a:100:42e5:9b3b]:4242
+   PKTDUMP: data received:
+   ~~ SNIP  0 - size:  13 byte, type: NETTYPE_UNDEF (0)
+   00000000  52  49  4F  54  20  54  65  73  74  69  6E  67  21
+   ~~ SNIP  1 - size:   8 byte, type: NETTYPE_UDP (4)
+      src-port:  4242  dst-port:  4242
+      length: 21  cksum: 0xee9e
+   ~~ SNIP  2 - size:  40 byte, type: NETTYPE_IPV6 (2)
+   traffic class: 0x00 (ECN: 0x0, DSCP: 0x00)
+   flow label: 0x00000
+   length: 21  next header: 17  hop limit: 64
+   source address: fe80::9c1a:100:42e5:9b3b
+   destination address: fe80::30a9:fa65:106b:1114
+   ~~ SNIP  3 - size:  24 byte, type: NETTYPE_NETIF (-1)
+   if_pid: 7  rssi: -43  lqi: 255
+   flags: 0x0
+   src_l2addr: 9E:1A:01:00:42:E5:9B:3B
+   dst_l2addr: 32:A9:FA:65:10:6B:11:14
+   ~~ PKT    -  4 snips, total size:  85 byte
+   ```


miri64:
On one hand it is good to add a Zephyr task specifically, on the other: isn't Zephyr not just using lwIP?

leandrolanzieri:
>  on the other: isn't Zephyr not just using lwIP?

Hmm I'm [not really sure](https://docs.zephyrproject.org/latest/guides/networking/net-stack-architecture.html) about that.

PeterKietzmann:
> Zephyr not just using lwIP?

Even if this is the case, then there is still MAC and driver interoperability

fjmolinas:
> > Zephyr not just using lwIP?
> 
> Even if this is the case, then there is still MAC and driver interoperability 

Yep, +1 for this change!

jia200x:
+1!

08-interop/08-interop.md
```diff
@@ -169,3 +169,109 @@ Task #10 - UDP between lwIP and emb6 on iotlab-m3
...
+node running Zephyr.
+
+### Testing procedure
+
+#### Setting up the Zephyr node
+
+1. Follow the
+['Getting Started'](https://docs.zephyrproject.org/latest/getting_started/index.html)
+guide on Zephyr's documentation.
+
```
miri64: Newline not needed (in fact it unnecessarily spreads out the numbered list.

08-interop/08-interop.md
```diff
@@ -169,3 +169,109 @@ Task #10 - UDP between lwIP and emb6 on iotlab-m3
...
+   PAN ID 43981 (0xabcd)
+
+   uart:~$ ieee802154 get_chan
+   Channel 26
+   ```
+
+
+#### Setting up the RIOT node
+
+1. Flash on the iotlab-m3 board the [gnrc_networking] example or the [gnrc_udp]
```
miri64:
```suggestion
1. Flash the [gnrc_networking] example or the [gnrc_udp] test to the iotlab-m3 board
```

08-interop/08-interop.md
```diff
@@ -169,3 +169,109 @@ Task #10 - UDP between lwIP and emb6 on iotlab-m3
...
+   test.
+
+2. Set the channel and PAN ID to the same values as the Zephyr node, running:
```
miri64:
```suggestion
2. Set the channel and PAN ID to the same values as the Zephyr node:
```

08-interop/08-interop.md
```diff
@@ -169,3 +169,109 @@ Task #10 - UDP between lwIP and emb6 on iotlab-m3
...
+   > ifconfig 6 set chan 26
+   ifconfig 6 set chan 26
+   success: set channel on interface 6 to 26
+
+   > ifconfig 6 set pan_id 0xabcd
+   ifconfig 6 set pan_id 0xabcd
+   success: set network identifier on interface 6 to 0xabcd
+   ```
+
+3. The UDP server application on the Zephyr node will be listening to port 4242.
```
miri64:
```suggestion
3. The UDP server application on the Zephyr node will be listening on port 4242.
```

08-interop/08-interop.md
```diff
@@ -169,3 +169,109 @@ Task #10 - UDP between lwIP and emb6 on iotlab-m3
...
+   Success: started UDP server on port 4242
+   ```
+
+4. Send packets to the echo server in the Zephyr node, they should be echoed and
+   printed on the RIOT side. Zephyr node will inform of the received packet as
+   well on the shell.
+
+   ```sh
+   # On iotlab-m3:riot node
+   > udp send fe80::d419:100:7ae4:9b3b 4242 "RIOT Testing!"
```
miri64: Maybe provide the expected output here.

08-interop/08-interop.md
```diff
@@ -169,3 +169,109 @@ Task #10 - UDP between lwIP and emb6 on iotlab-m3
...
+   ifconfig 6 set chan 26
+   success: set channel on interface 6 to 26
+
+   > ifconfig 6 set pan_id 0xabcd
+   ifconfig 6 set pan_id 0xabcd
+   success: set network identifier on interface 6 to 0xabcd
+   ```
+
+3. The UDP server application on the Zephyr node will be listening to port 4242.
+   Start the UDP server on the same port on the RIOT node:
```
miri64: Does the echo server reply to that port?

leandrolanzieri: Yes, it echoes to the same port it is listening on.

leandrolanzieri:
Here is the output when running the task:

**RIOT node**
```
> udp server start 4242
2020-03-20 09:13:46,566 #  udp server start 4242
2020-03-20 09:13:46,568 # Success: started UDP server on port 4242
> udp send fe80::9c1a:100:42e5:9b3b 4242 "This is RIOT!"
2020-03-20 09:14:01,350 #  udp send fe80::9c1a:100:42e5:9b3b 4242 "This is RIOT!"
2020-03-20 09:14:01,353 # Success: sent 13 byte(s) to [fe80::9c1a:100:42e5:9b3b]:4242
2020-03-20 09:14:01,365 # PKTDUMP: data received:
2020-03-20 09:14:01,368 # ~~ SNIP  0 - size:  13 byte, type: NETTYPE_UNDEF (0)
2020-03-20 09:14:01,370 # 00000000  54  68  69  73  20  69  73  20  52  49  4F  54  21
2020-03-20 09:14:01,373 # ~~ SNIP  1 - size:   8 byte, type: NETTYPE_UDP (4)
2020-03-20 09:14:01,375 #    src-port:  4242  dst-port:  4242
2020-03-20 09:14:01,377 #    length: 21  cksum: 0x05d2
2020-03-20 09:14:01,381 # ~~ SNIP  2 - size:  40 byte, type: NETTYPE_IPV6 (2)
2020-03-20 09:14:01,384 # traffic class: 0x00 (ECN: 0x0, DSCP: 0x00)
2020-03-20 09:14:01,387 # flow label: 0x00000
2020-03-20 09:14:01,389 # length: 21  next header: 17  hop limit: 64
2020-03-20 09:14:01,391 # source address: fe80::9c1a:100:42e5:9b3b
2020-03-20 09:14:01,394 # destination address: fe80::30a9:fa65:106b:1114
2020-03-20 09:14:01,396 # ~~ SNIP  3 - size:  24 byte, type: NETTYPE_NETIF (-1)
2020-03-20 09:14:01,397 # if_pid: 7  rssi: -43  lqi: 255
2020-03-20 09:14:01,398 # flags: 0x0
2020-03-20 09:14:01,398 # src_l2addr: 9E:1A:01:00:42:E5:9B:3B
2020-03-20 09:14:01,399 # dst_l2addr: 32:A9:FA:65:10:6B:11:14
2020-03-20 09:14:01,401 # ~~ PKT    -  4 snips, total size:  85 byte
```
**Zephyr node**
```
[00:02:00.019,000] <dbg> net_echo_server_sample.process_udp: (0x2000049c): UDP (IPv6): Received and replied with 13 bytes
```

And the wireshark:
image

08-interop/08-interop.md
leandrolanzieri: Removed newlines

08-interop/08-interop.md
leandrolanzieri: Added output

08-interop/08-interop.md
miri64: That's weird... why not to the source port?

leandrolanzieri: Sorry, I confused myself with another example I was reading. I checked [the code](https://github.com/zephyrproject-rtos/zephyr/blob/ba0aea3cd3f4aee1f21d3e1b798e6713827465f3/samples/net/sockets/echo_server/src/udp.c#L118-L119) and it sends back to the address as is, so it should be the source port.

miri64: Ah, but we use the same source port as the destination port for `udp send`, right! Still, something like this should be mentioned here in my opinion to avoid confusion.

leandrolanzieri: Changed to clarify this point.

leandrolanzieri:
@miri64  Thanks for reviewing

08-interop/08-interop.md
```diff
@@ -169,3 +169,123 @@ Task #10 - UDP between lwIP and emb6 on iotlab-m3
...
+   uart:~$ ieee802154 get_pan_id
+   PAN ID 43981 (0xabcd)
+
+   uart:~$ ieee802154 get_chan
+   Channel 26
+   ```
+
+#### Setting up the RIOT node
+
+1. Flash the [gnrc_networking] example or the [gnrc_udp] test to the iotlab-m3 board
```
PeterKietzmann: Why not be specific about the test application?

08-interop/08-interop.md
```diff
@@ -169,3 +169,123 @@ Task #10 - UDP between lwIP and emb6 on iotlab-m3
...
+   test.
+2. Set the channel and PAN ID to the same values as the Zephyr node:
+   ```sh
+   # On iotlab-m3:riot node
+   > ifconfig 6 set chan 26
```
PeterKietzmann: With *gnrc_networking* the interface number is 7.

08-interop/08-interop.md
```diff
@@ -169,3 +169,123 @@ Task #10 - UDP between lwIP and emb6 on iotlab-m3
...
+   udp server start 4242
+   Success: started UDP server on port 4242
+   ```
+4. Send packets to the echo server in the Zephyr node, they should be echoed and
+   printed on the RIOT side. Zephyr node will inform of the received packet as
+   well on the shell.
+
+   ```sh
+   # On iotlab-m3:riot node
+   > udp send fe80::d419:100:7ae4:9b3b 4242 "RIOT Testing!"
```
PeterKietzmann: This address may change so shouldn't we rather put something like ´<IPv6 addr>` here?

miri64:
> How about we simply provide a binary file for release testing? This simplifies the testing process **a lot** and makes it reproducible.

This would either quickly lead to Zephyr and RIOT to go out of sync (or require some automation or maintainer work on our side to provide that binary in a regular manner).

08-interop/08-interop.md
miri64: Both _should_ be the same basically. The only difference is that `gnrc_udp` expects a number for the payload (representing the payload length) and `gnrc_networking` a string (representing the actual payload).

08-interop/08-interop.md
miri64: This should be solved once https://github.com/RIOT-OS/RIOT/pull/12738 and we adapt the `ifconfig` shell command for it (which due to the test scripts might be a quite massive change). :sweat_smile:

08-interop/08-interop.md
miri64: We use example addresses all the time... But maybe to make it more clearer this indeed should be `<target link-local address>` or something like that.

08-interop/08-interop.md
PeterKietzmann: If it should be the same, then I don't see the difference in selecting one of both. If there is a relevant  difference, then we should consider testing both.

PeterKietzmann:
> This would either quickly lead to Zephyr and RIOT to go out of sync (or require some automation or maintainer work on our side to provide that binary in a regular manner).

For sure, the binary should be updated every once in a while. However, I don't see a big problem if this doesn't happen every (RIOT) release. We could still say "passes interop tests against zephyr Vxyz". My point is that I see a risk this test becomes a single point of failure like the contiki test. Only very few people where able to run it, and if these people were not available, we somehow skipped it. "Debugging" up to date zephyr shouldn't be the responsibility of the release manager, should it? As explained yesterday, the mentioned application doesn't build for the target platform on zephyrs current master. How do we deal with that?

08-interop/08-interop.md
miri64: > If it should be the same, then I don't see the difference in selecting one of both. If there is a relevant difference, then we should consider testing both.

The only reason `gnrc_udp` exists in the first place is because the release specs were asking for large payloads that were not doable with a string input and 64 byte shell buffer ;-). So when it comes to release specs one should prefer `gnrc_udp`.

benpicco:
If you disable the shell history and reduce the size of the network buffers in Kconfig (`make menuconfig`) the example should still fit on the `samr21-xpro` - I can try to provide a minimal config.

leandrolanzieri:
Regarding fitting the board, we could also consider some other board (like the [nRF52840 DK](https://docs.zephyrproject.org/latest/boards/arm/nrf52840dk_nrf52840/doc/index.html)) which could fit the example better.

I just tried compiling the example for the `samr21-xpro` on current Zephyr master and it fails:
```
In file included from ../soc/arm/atmel_sam0/samr21/soc.h:29,
                 from ../include/arch/arm/aarch32/cortex_m/cmsis.h:17,
                 from ../arch/arm/include/aarch32/cortex_m/stack.h:23,
                 from ../arch/arm/include/kernel_arch_data.h:33,
                 from /home/leandro/Work/git/zephyrproject/zephyr/arch/arm/core/offsets/offsets_aarch32.c:26,
                 from /home/leandro/Work/git/zephyrproject/zephyr/arch/arm/core/offsets/offsets.c:12:
/home/leandro/Work/git/zephyrproject/modules/hal/atmel/asf/sam0/include/samr21/samr21g18a.h:230:10: fatal error: core_cm0plus.h: No such file or directory
  230 | #include <core_cm0plus.h>
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
ninja: build stopped: subcommand failed.
FATAL ERROR: command exited with status 1: /usr/local/bin/cmake --build /home/leandro/Work/git/zephyrproject/zephyr/build
```
Perhaps it would be a good idea to point to releases for this test.

PR #157, opened 2020-07-11
08-interop: deprecate 9-10

### LINKED ISSUE #141: RFC: Task 8: Remove emb6 interop tests
With emb6 being deprecated since https://github.com/RIOT-OS/RIOT/pull/12389 I'm wondering if it still makes sense to include emb6 in the release testing until 2020.06 or not.

miri64:
`emb6` was deprecated in https://github.com/RIOT-OS/RIOT/pull/12389 and removed in https://github.com/RIOT-OS/RIOT/pull/14494 so it does not make much sense to test this anymore, even for 2020.07

Resolves #141

### DIFF
--- a/08-interop/08-interop.md
+++ b/08-interop/08-interop.md
@@ -154,21 +154,7 @@ Link-local UDP over IPv6 packet exchange (payload length 8) between an iotlab-m3
 node running RIOT with GNRC and an iotlab-m3 node running RIOT with lwIP (in
 both directions).
 
-Task #09 - UDP between GNRC and emb6 on iotlab-m3
-=================================================
-### Description
-
-Link-local UDP over IPv6 packet exchange (payload length 8) between an iotlab-m3
-node running RIOT with GNRC and an iotlab-m3 node running RIOT with emb6.
-
-
-Task #10 - UDP between lwIP and emb6 on iotlab-m3
-=================================================
-### Description
-
-Link-local UDP over IPv6 packet exchange (payload length 8) between an iotlab-m3
-node running RIOT with lwIP and an iotlab-m3 node running RIOT with emb6 (in
-both directions).
+Task #09-#10 deprecated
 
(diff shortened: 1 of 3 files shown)


leandrolanzieri:
It makes sense to keep the numbering of the tasks for now, but maybe we should discuss for how long, and maybe come up with a removal procedure for tasks.

On one side keeping numbering reduces confusion when looking at previous release proceses, but on the other side it forces us to reserve those empty tasks numbers. Would something similar to the `LOSTANDFOUND.md` work here?

miri64:
> On one side keeping numbering reduces confusion when looking at previous release proceses, but on the other side it forces us to reserve those empty tasks numbers. Would something similar to the `LOSTANDFOUND.md` work here?

Only if we keep those numbers. If we re-use them `LOSTANDFOUND.md` doesn't make much sense, as people would look at it to search for a missing number.

How about keeping them reserved for at least a year and then documenting the replacement in a `REPLACED.md`?

miri64:
Or just as a section in the top-level `README`. I don't expect this to happen very often so the list wouldn't be that long.

leandrolanzieri:
> How about keeping them reserved for at least a year and then documenting the replacement in a `REPLACED.md`?

Sound good. I think it does not hurt to have the extra file. It would be nice also to specify the first and last release the task has been valid and the specific task number it used to be. So one could trace back depending on which release tests are being inspected.

miri64:
Like [this](https://github.com/RIOT-OS/Release-Specs/pull/157/commits/a176f2763d7fd43c8f358a97d66a11f6f9e0868a)?

leandrolanzieri:
> Like [this](https://github.com/RIOT-OS/Release-Specs/blob/198e46c50c1d7dfd1361ff9aa0fff4e912325444/REPLACED.md)?

Looks good. Please squash

miri64:
Squashed

miri64:
Thanks!

PR #155, opened 2020-07-10
testutils: Automated testing environment for release tests using pytest

miri64:
### Contribution description

Third time's a charm: after https://github.com/RIOT-OS/Release-Specs/pull/79 and some major overhaul of that framework to use (then) `riotnode` (now `riotctrl`) and `pytest` by @fjmolinas, I ported the latter to run with the new [`riotctrl`](https://github.com/RIOT-OS/riotctrl/) python package and the in-tree [`riotctrl_shell`](https://github.com/RIOT-OS/RIOT/tree/b05b2fddb967b170f2c2cdcfda5641085d55fbcc/dist/pythonlibs/riotctrl_shell) module.

I also cleaned up the by now very complicated history of that effort and squashed a lot of commits. Since the most of the porting work was done by @fjmolinas (I just moved some code around and enhanced his code) I made him the main author of the work in the commit history, but added all co-authors to the respective commits for the changes they made.

### Testing procedure
Run `tox` and the tests will be run (there are none at the moment, see #160, #161, #162, #163, #158, and #159).
Check-out #164 to run all tests.
There is also some more info in the README how to select specific groups of tests.

Self-tests are run by Github Actions and can be run locally with `tox -- --self-test`.

### Issues/PRs references
Alternative to #79.

Required for #160, #161, #162, #163, #158, and #159

PR #158, opened 2020-07-13
07-multi-hop: port automated tests to pytest 

miri64:
This ports the existing automated tests for spec 7 to the new, `pytest`-based framework.

Requires #155


miri64:
Rebased to current #155 (for which I removed the spec tests)

miri64:
Rebased to current master.

miri64:
Rebased to current master.

miri64:
Rebased to current master.

07-multi-hop/test_spec07.py
```diff
@@ -0,0 +1,240 @@
...
+        nodes_netifs.append({
+            "netif": key,
+            "hwaddr": netifs[key]["long_hwaddr"],
+            "lladdr": [addr["addr"] for addr in netifs[key]["ipv6_addrs"] if
+                       addr["scope"] == "link"][0],
+        })
+    return nodes_netifs
+
+
+# pylint: disable=W0621
```
fjmolinas, line 67: Not needed.
```suggestion
```

07-multi-hop/test_spec07.py
```diff
@@ -0,0 +1,240 @@
...
+            # set to-route
+            node.ifconfig_l2filter_add(nodes_netifs[i]["netif"],
+                                       nodes_netifs[i + 1]["hwaddr"])
+        if i > 0:
+            # set from-route
+            node.ifconfig_l2filter_add(nodes_netifs[i]["netif"],
+                                       nodes_netifs[i - 1]["hwaddr"])
+
+
+# pylint: disable=W0621
```
fjmolinas, line 80:
```suggestion
```

miri64:
>  can we open an issue for this instead? (same for `l2filter` stuff).

Huh? `l2filter` is in master already.

07-multi-hop/test_spec07.py
miri64: Reasoning?

miri64:
>  there is a todo mention, can we open an issue for this instead?

Why not both? I think it's good to keep the reminder in the code as well.

07-multi-hop/test_spec07.py
fjmolinas: Pylint didn't complain when ran locally? Did I do it wrong?

fjmolinas:
> Why not both? I think it's good to keep the reminder in the code as well.

Wrong choice of words, both is perfect.

miri64:
> […] can we open an issue for this […]?

See https://github.com/RIOT-OS/RIOT/issues/14637

07-multi-hop/test_spec07.py
miri64: Ah, you are right! `nodes_netifs` is used in those functions after all!

miri64:
> Also seems like `_get_global_addr` and `_get_local_addr` could also be put in common somewhere.

You mean `_get_root_addr` with the second? There is no `_get_local_addr`.

miri64:
> You mean `_get_root_addr` with the second? There is no `_get_local_addr`.

In that case I at least disagree for that one as this is very RPL specific and the only RPL tests are in this spec.

miri64:
For global address I agree.

fjmolinas:
> You mean _get_root_addr with the second? There is no _get_local_addr.

Sorry I think at the same time I was looking at the `native` `get_link_local`, and I meant it in the sense that both could exist for non-native.

miri64:
> Sorry I think at the same time I was looking at the `native` `get_link_local`, and I meant it in the sense that both could exist for non-native.

There are no `native` tests here???!? But regarding `global_addr` there is a change in the making, I just run the tests for it now. Let's see how you like it :-).

fjmolinas:
>   There are no native tests here???!? But regarding global_addr there is a change in the making, I just run the tests for it now. Let's see how you like it :-).

Yes I know I was looking at the `testutils` directory, there there are some native functions, and I saw `get_link_local`. Sorry for not being clear about that.

miri64:
Have a look at the new integration of `global_addr` to `testutils`.

fjmolinas:
Looks good please squash and rebase @miri64!

miri64:
Done

PR #159, opened 2020-07-13
08-interop: initial import of automated tests

miri64:
This provides automated tests (where possible with the current framework) for spec 8 to the new, `pytest`-based framework.

Requires #155

### DIFF
--- a/08-interop/test_spec08.py
+++ b/08-interop/test_spec08.py
@@ -1,0 +1,88 @@
+import re
+import sys
+
+import pexpect.replwrap
+import pytest
+
+import riotctrl.shell
+from riotctrl_shell.gnrc import GNRCICMPv6Echo
+from riotctrl_shell.netif import Ifconfig
+
+from testutils.native import bridge, get_link_local, get_ping_cmd, \
+                             interface_exists
+from testutils.shell import lladdr, ping6, GNRCUDP
+
+
+GNRC_APP = 'examples/gnrc_networking'
+LWIP_APP = 'tests/lwip'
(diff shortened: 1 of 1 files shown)


miri64:
Rebased to current #155 (for which I removed the spec tests)

miri64:
Rebased and squashed to current master

miri64:
Rebased to current master

miri64:
Rebased to current master

08-interop/test_spec08.py
```diff
@@ -0,0 +1,88 @@
...
+                         [pytest.param(['native'])],
+                         indirect=['nodes'])
+def test_task01(riot_ctrl, log_nodes):
+    node = riot_ctrl(0, GNRC_APP, Shell, port='tap0')
+    linux = pexpect.replwrap.bash()
+
+    node_iface, node_addr = lladdr(node.ifconfig_list())
+    assert node_addr.startswith("fe80::")
+    linux_iface = bridge('tap0')
+    linux_addr = get_link_local(linux_iface)
```
fjmolinas, line 38: Maybe when the other tasks get implemented this will make sense in a common module

miri64: I don't understand? Those tasks would also be implemented also in this script, so no need to move that function to a separate `common` module.

miri64:
> I think task 6 7 could also be done

Yeah, but `@only_local` ;-).

08-interop/test_spec08.py
fjmolinas: ah true. nvm

PR #160, opened 2020-07-13
03-single-hop-ipv6-icmp: initial import of automated tests

miri64:
Taken out of the original #155, this provides automated tests (where possible with the current framework) for spec 3 to the new, pytest-based framework.

Requires #155

### DIFF
--- a/03-single-hop-ipv6-icmp/test_spec03.py
+++ b/03-single-hop-ipv6-icmp/test_spec03.py
@@ -1,0 +1,182 @@
+import sys
+import time
+
+import pexpect.replwrap
+import pytest
+
+from riotctrl_shell.gnrc import GNRCICMPv6Echo, GNRCPktbufStats
+from riotctrl_shell.netif import Ifconfig
+
+from testutils.asyncio import wait_for_futures, timeout_futures
+from testutils.native import bridged, bridge, get_ping_cmd, interface_exists
+from testutils.shell import ping6, pktbuf, lladdr
+
+
+APP = 'tests/gnrc_udp'
+TASK05_NODES = 11
+pytestmark = pytest.mark.rc_only()
+
+
+class Shell(Ifconfig, GNRCICMPv6Echo, GNRCPktbufStats):
+    pass
+
+
+@pytest.mark.skipif(not bridged(["tap0", "tap1"]),
+                    reason="tap0 and tap1 not bridged")
+# nodes passed to riot_ctrl fixture
+@pytest.mark.parametrize('nodes',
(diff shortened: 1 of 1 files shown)


miri64:
Rebased and squashed to current master.

03-single-hop-ipv6-icmp/test_spec03.py
```diff
@@ -0,0 +1,182 @@
...
+                                         pinged_addr=pinged_addr),
+                                 async_=True, timeout=60 * 60)
+        futures.append(out)
+    timeout_futures(futures, 60 * 60)
+
+    time.sleep(60)
+    assert pktbuf(node).is_empty()
+
+
+@pytest.mark.xfail(reason="See https://github.com/RIOT-OS/RIOT/issues/12565")
```
fjmolinas, line 123: I think in this case `xfail` is valid.

miri64:
Rebased and squashed to current master

miri64:
Rebased to current master.

03-single-hop-ipv6-icmp/test_spec03.py
```diff
@@ -0,0 +1,182 @@
...
+    pinger, pinged = (
+        riot_ctrl(0, APP, Shell, port='tap0'),
+        riot_ctrl(1, APP, Shell, port='tap1'),
+    )
+
+    _, pinged_addr = lladdr(pinged.ifconfig_list())
+    assert pinged_addr.startswith("fe80::")
+
+    res = ping6(pinger, pinged_addr,
+                count=1000, interval=100, packet_size=2000)
```
fjmolinas, line 174:
```suggestion
                count=1000, interval=100, packet_size=2048)
```

03-single-hop-ipv6-icmp/test_spec03.py
fjmolinas, line 123: Shouldn't this test require `ping -f` since the interval is 0?

```suggestion
@pytest.mark.sudo_only      # ping -f requires root
@pytest.mark.xfail(reason="See https://github.com/RIOT-OS/RIOT/issues/12565")
```

miri64: No, this task is multiple `native` nodes pinging one `native` node, not the host force pinging one `native` node (that's task 04). Also the reason why this test requires 11 TAP interfaces ;-).

03-single-hop-ipv6-icmp/test_spec03.py
miri64: Addressed and amended directly

PR #165, opened 2020-07-13
10-icmpv6-error: add task to test multicast destination

### LINKED ISSUE #136: Task 10: add test case for multicast packets
There was a bug in `gnrc_icmpv6_error` found and fixed in https://github.com/RIOT-OS/RIOT/pull/12175. A test case should be added to assure that no regression may happen.

miri64:
Resolves #136.

### DIFF
--- a/10-icmpv6-error/10-icmpv6-error.md
+++ b/10-icmpv6-error/10-icmpv6-error.md
@@ -329,4 +329,25 @@ headers as is possible within the given MTU.
 \>96% of the echo requests should be replied. There should only be at most one
 echo reply to each echo request.
 
+Task #11 - Multicast destination address in triggering packet
+=============================================================
+### Description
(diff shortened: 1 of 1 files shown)


miri64:
I forgot the [rocessing rules for IPv6 messages](https://tools.ietf.org/html/rfc4443#section-2.4). The proposed task will not work as intended, as multicast destinations only cause the emission of certain error types. Destination unreachable is not one of them.

miri64:
The only error we support, that supports emission on multicast is "Packet to Big", however for that we need multicast forwarding for which we know there are problems in GNRC. See https://github.com/RIOT-OS/RIOT/issues/4527 and https://github.com/RIOT-OS/RIOT/issues/5230. So I close this for now. We have the issue as a reminder.

PR #171, opened 2020-07-16
pytest: fail test on flash fail

miri64:
While debugging a problem using the new framework I noticed, that it can be really annoying, when the flash fails and the test just runs with the old image ;-).

This makes a test fail if the `make flash` fails for some reason (e.g. build error) by using the `check=True` parameter. Likewise, it makes piping the output to `DEVNULL` optional (bound to `--hide-output`), to debug potential problems during the build. Lastly, it now uses the `DEVNULL` of `subprocess` instead of the self-defined one.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -7,6 +7,7 @@ See https://docs.pytest.org/en/stable/fixture.html#conftest-py-sharing-fixture-f
 """     # noqa: E501
 
 import os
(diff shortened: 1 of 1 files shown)


miri64:
Rebase to current master.

miri64:
Rebased.

PR #173, opened 2020-07-17
testutils: mark tests in RC tracking issue and comment on status

miri64:
This implements the functionality seen in https://github.com/RIOT-OS/Release-Specs/issues/172, so marking a task as done in the tracking issue if it is succeeding (and not already marked as done) and commenting the results (if they are not already provided)


miri64:
Rebased to current master

conftest.py
```diff
@@ -92,6 +93,13 @@ def pytest_collection_modifyitems(config, items):
             item.add_marker(rc_only_mark)
 
 
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_logreport(report):
+    result = yield
+    testutils.github.update_issue(report)
+    return result
```
fjmolinas, line 100: I don't understand why `return yield`?

miri64: `yield` in a keyword to yield to the caller. This can produce results in fact. So what is here is

```py
return result
```

not `return yield`.

This is done in case the wrapping `pytest_runtest_logreport()` returns something. See also https://docs.pytest.org/en/stable/writing_plugins.html#hookwrapper-executing-around-other-hooks

miri64: Then again: It is not expected to return anything... so no idea... will remove.

fjmolinas: Thanks for the pointers to the documentation, it is clearer to me know.

fjmolinas:
please rebase and squash @miri64

miri64:
Rebased and squashed.

PR #178, opened 2020-07-28
pytest: add workaround to add USEMODULE with BUILD_IN_DOCKER

miri64:
... And set `tox` environment for GitHub actions

This is the necessary part from #174 that is needed weather the workflow will be added here or to RIOT/RIOT-OS.


miri64:
Rebased to current master

tox.ini
```diff
@@ -16,6 +16,11 @@ commands =
 [testenv:test]
 passenv =
   BUILD_IN_DOCKER
+  DOCKER_ENV_VARS
+  DOCKER_MAKE_ARGS
+  GITHUB_REPOSITORY
+  GITHUB_RUN_ID
+  GITHUB_SERVER_URL
```
fjmolinas, line 23: Why are these added?

conftest.py
```diff
@@ -193,6 +193,12 @@ def update_env(node, modules=None, cflags=None, port=None, termflags=None):
         node.env['USEMODULE'] = ' '.join(str(m) for m in modules)
     elif modules is not None:
         node.env['USEMODULE'] = modules
+    if "USEMODULE" in node.env and \
+       os.environ.get('BUILD_IN_DOCKER', 0) == '1':
+        # workaround to inject USEMODULE into docker container
+        # see: https://github.com/RIOT-OS/RIOT/issues/14504
+        node.env['DOCKER_ENVIRONMENT_CMDLINE'] = "-e 'USEMODULE={}'" \
+            .format(node.env['USEMODULE'])
```
fjmolinas, line 201: Mimics current handling done when launching from command line, not ideal, not intuitive but its the current way to do it, maybe it can be eventually fixed in master and this patch removed, but for now its the way to go

tox.ini
miri64: Oops, those belong to #173. Must be a leftover from when I took this out of #174.

conftest.py
miri64: I just realized: `DOCKER_ENVIRONMENT_CMDLINE` might be already set with something by the user. Will change, so it is amended if set.

tox.ini
miri64: Moved to #173

conftest.py
miri64: Done.

tox.ini
```diff
@@ -16,6 +16,8 @@ commands =
 [testenv:test]
 passenv =
   BUILD_IN_DOCKER
+  DOCKER_ENV_VARS
+  DOCKER_MAKE_ARGS
```
fjmolinas, line 20: I still don't get why these are passed, it seems to be realated to https://github.com/RIOT-OS/RIOT/issues/14504, but the current solution is not using these variables.

miri64: `tox` wipes the environment to a minimal set. So if they are not provided they are not set to `docker`. Just try it out yourself using `tox -- --capture=tee-sys` with the environment variables provided in #14646, e.g.:

- https://github.com/RIOT-OS/RIOT/pull/14646/files#diff-64da09ece1cfe234928d5e51cfca3e09R177 and
- https://github.com/RIOT-OS/RIOT/pull/14646/files#diff-64da09ece1cfe234928d5e51cfca3e09R28

With those environment variables exposed to tox:

```
$ BUILD_IN_DOCKER=1 DOCKER_MAKE_ARGS=-j8 tox  -- -k "spec03 and task01" --non-RC --capture=tee-sys
[…]
03-single-hop-ipv6-icmp/test_spec03.py::test_task01[nodes0] Launching build container using image "riot/riotbuild:latest".
docker run --rm --tty --user $(id -u) -v '/usr/share/zoneinfo/Europe/Berlin:/etc/localtime:ro' -v '/home/mlenders/Repositories/RIOT-OS/RIOT:/data/riotbuild/riotbase:delegated' -e 'RIOTBASE=/data/riotbuild/riotbase' -e 'CCACHE_BASEDIR=/data/riotbuild/riotbase' -e 'BUILD_DIR=/data/riotbuild/riotbase/build' -e 'RIOTPROJECT=/data/riotbuild/riotbase' -e 'RIOTCPU=/data/riotbuild/riotbase/cpu' -e 'RIOTBOARD=/data/riotbuild/riotbase/boards' -e 'RIOTMAKE=/data/riotbuild/riotbase/makefiles'      -e 'BOARD=native'  -w '/data/riotbuild/riotbase/tests/gnrc_udp/' 'riot/riotbuild:latest' make    -j8 
Building application "tests_gnrc_udp" for "native" with MCU "native".
[…]
```

without these variables exposed:

```
$ git diff
diff --git a/tox.ini b/tox.ini
index ee60cfd..0c9a579 100644
--- a/tox.ini
+++ b/tox.ini
@@ -15,9 +15,6 @@ commands =
 
 [testenv:test]
 passenv =
-  BUILD_IN_DOCKER
-  DOCKER_ENV_VARS
-  DOCKER_MAKE_ARGS
   HOME
   IOTLAB_SITE
   RIOTBASE
$ BUILD_IN_DOCKER=1 DOCKER_MAKE_ARGS=-j8 tox  -- -k "spec03 and task01" --non-RC --capture=tee-sys
03-single-hop-ipv6-icmp/test_spec03.py::test_task01[nodes0] Building application "tests_gnrc_udp" for "native" with MCU "native".
```

miri64: > I still don't get why these are passed, it seems to be realated to [RIOT-OS/RIOT#14504](https://github.com/RIOT-OS/RIOT/issues/14504), but the current solution is not using these variables.

The current solution would not build properly in https://github.com/RIOT-OS/RIOT/pull/14646 btw. So maybe they could be used to their own PR or I can split them up in a separate PR, either way: you decide.

fjmolinas: Thank for the explanation, now I understand. No need to have another PR for this, but I would like a different commit, and I think it makes sense to add `DOCKER_ENVIRONMENT_CMDLINE` as well to the list.

miri64: Done. I will move them to a different commit once I squash

fjmolinas: please squash.

miri64: Squashed

miri64:
Thanks for the review, btw!

PR #181, opened 2020-07-31
04-single-hop-6lowpan-icmp/task04: reconnect terminal after long test

miri64:
Task 4.4 fails on the CI, since the Github servers appear to silently drop the TCP connection to the node after a while (the pinged node is not accessed for the duration of the ping). By just starting the terminal anew, this is fixed. I will start a new run, once I merged this PR into #174.

This requires https://github.com/RIOT-OS/riotctrl/pull/6 to work.

PR #183, opened 2020-08-01
04-single-hop-6lowpan-icmp/task04: xfail pktbuf on pinged node

miri64:
The lazy version for "fixing" 4.4 if we can not resolve https://github.com/RIOT-OS/Release-Specs/pull/181#issuecomment-667517820.


miri64:
Rebased to current master

miri64:
Rebased to current master

PR #187, opened 2020-08-05
tox: remove unneeded gh-actions env

miri64:
See https://github.com/RIOT-OS/RIOT/pull/14646#pullrequestreview-461640916.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,10 +2,6 @@
 envlist = test,flake8,pylint
 skipsdist = True
 
-[gh-actions]
-python =
-    3.8: py38,flake8,pylint,test
-
 [testenv]
 basepython = python3
 commands =


miri64:
See https://github.com/miri64/RIOT/actions/runs/196341029 for a release test with these changes included.

miri64:
> ACK (Assuming the iotlab task failure is unrelated)

Should be. I don't se a reason, why removing this should cause the `ifconfig` run into an assertion. I encountered this bug before and usually the only thing helping is flashing an older version of RIOT (I use `2019.10` currently) and doing a reset. Because once it happens it is stuck there. Seems to be more an hardware issue in the `m3` nodes in the testbed it seems.

miri64:
Rebased to current master.

PR #188, opened 2020-08-05
pytest: use records to store RIOT version and experiment ID of test

miri64:
I noticed that I had this still lying around from before my vacation. This should make it easier to track after the fact which

1. RIOT version was used for a given test and
2. Which IoT-LAB experiment a test ran in.


miri64:
(this is based on an offline suggestion by @MrKevinWeiss)

miri64:
Mh after re-running the tests I get warnings

```
03-single-hop-ipv6-icmp/test_spec03.py::test_task01[nodes0]
  03-single-hop-ipv6-icmp/test_spec03.py:25: PytestWarning: record_property is incompatible with junit_family 'xunit2' (use 'legacy' or 'xunit1')
    @pytest.mark.skipif(not bridged(["tap0", "tap1"]),

04-single-hop-6lowpan-icmp/test_spec04.py::test_task01[nodes0]
  04-single-hop-6lowpan-icmp/test_spec04.py:22: PytestWarning: record_property is incompatible with junit_family 'xunit2' (use 'legacy' or 'xunit1')
    @pytest.mark.iotlab_creds

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

@MrKevinWeiss do you know of an `xunit2` compliant way to do this?

miri64:
Rebased to current master and also added the Release-Specs version as a property.

MrKevinWeiss:
> @MrKevinWeiss do you know of an xunit2 compliant way to do this?

I don't know about this but I like the idea of the PR.

MrKevinWeiss:
Do we now have a way to get this info or should we continue with this?

miri64:
> Do we now have a way to get this info or should we continue with this?

If we can live with the warning, I guess we can continue. Alternatively, we of course also could choose a different output format than `xunit2` (which according to my last google search on that topic seems to be pytest specific anyways?)

MrKevinWeiss:
Or we could just use `junit_family=legacy`, was there a reason that we went with xunit2?

MrKevinWeiss:
I will give it a try and see what breaks.

miri64:
> Or we could just use `junit_family=legacy`, was there a reason that we went with xunit2?

I think there were warnings to use `xunit2` ^^.

PR #190, opened 2020-08-06
09-coap/task01: use upstream CordEp ShellInteraction

pokgak:
This PR replaces the implementation of Cord EP ShellInteraction here with upstream as requested https://github.com/RIOT-OS/RIOT/pull/14715#issuecomment-669884370.

### Testing Procedure

```bash
sudo RIOTBASE=$(readlink -f ../RIOT) tox -e test -- -k "spec09 and task01" --non-RC
```

PR #191, opened 2020-08-06
09-coap/task01: fix typo

miri64:
I found a typo in task 9.1 while reviewing and testing #190. This never made any problems so far, since the `examples/cord_ep` application is configured as expected and more of a fail-safe than anything else.


miri64:
Rebased to current master.

aabadie (approved):
Obviously that's a typo.

ACK

PR #194, opened 2020-10-27
08-interop: gnrc_ipv6_nib_dns is required to have name resolution in the WPAN

benpicco:
`gnrc_ipv6_nib_dns` was split off `sock_dns`, but we need both for name resolution.


jia200x:
seems legit

jia200x (approved):
ACK

PR #200, opened 2021-01-12
gh-actions: don't use tox-gh-actions plugin

miri64:
@aabadie warned me to use that plugin, and now it shows... for some reason it does not execute neither flake8 nor pylint. This removes the usage of the `tex-gh-actions` plugin and also adds `3.9` tests.

### DIFF
--- a/.github/workflows/self-test.yml
+++ b/.github/workflows/self-test.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.8]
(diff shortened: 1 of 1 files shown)


jia200x:
I'm fine with the change. @aabadie are you also ok with it?

jia200x (approved):
ACK. For some reason this prevents running  flake8` and `pylint`. It would be a nice if at some point this can be done with such a plugin, but until we don't find a solution let's keep it as it's proposed here

PR #203, opened 2021-01-18
04-single-hop-6lowpan-icmp: add missing iotlab_creds

jia200x:
This PR adds some missing `@iotlab_creds` decorators to tasks 12 and 13. Without these, the release-tests won't pick them.


miri64:
> Without these, the release-tests won't pick them.

Actually, they _will_ pick them up, but also for runs where no credentials are provided ;-).

miri64 (approved):
ACK. Without this fix the tests are trying to be run without iot-lab credentials provided, resulting in unauthorized access errors, such as [these](https://github.com/RIOT-OS/RIOT/runs/1708631603?check_suite_focus=true#step:12:140)

PR #205, opened 2021-01-20
task7: update example_test_guide

MrKevinWeiss:
# CHANGES

- The default interface is 5 now
- The packet buffer check command is `pktbuf`
- There is a separate build command from the flash and term (because concurrency issues)
- A step to initialize the `rpl` instance is added for all nodes
- The experimental has the current state of the results

### DIFF
--- a/07-multi-hop/example_test_guide.md
+++ b/07-multi-hop/example_test_guide.md
@@ -7,19 +7,20 @@ node1 <-> node2 <-> node3 <-> node4_
 1. Book 4 [iotlab](https://www.iot-lab.info/testbed/dashboard) M3 nodes
 
 # Task 1 and 2 Setup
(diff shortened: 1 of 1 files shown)


jia200x:
thanks for the fix!
@miri64 does it look ok for you?

miri64:
I never used this, but looks alright.

MrKevinWeiss:
> I never used this, but looks alright.

For those who only deal with networking every release cycle :smile:

jia200x:
> I was just stating, that I was unsure why @jia200x explicitly asked me to have a look ;-).

I haven't either, but I thought you had :)

jia200x (approved):
ACK then :)

jia200x:
thanks @MrKevinWeiss !

PR #209, opened 2021-02-08
testutils/shell: add check_pktbuf

jia200x:
## Contribution description

This commit fixes some race conditions when checking that all pinged nodes
have an empty pktbuf. This PR adds an utility to wait for some seconds before probing the pktbuf status.

Without this commit some of these tests fail randomly (e.g see https://github.com/RIOT-OS/RIOT/actions/runs/537422117)
## Testing Procedure
```
RIOTBASE=<absolute_path_to_RIOT_folder> IOTLAB_SITE=grenoble tox -- --non-RC -k "spec04"
```
I'm explicitly choosing a different IoTLAB site because on 8/2/2021 all Saclay nodes are reserved.

### DIFF
--- a/04-single-hop-6lowpan-icmp/test_spec04.py
+++ b/04-single-hop-6lowpan-icmp/test_spec04.py
@@ -1,5 +1,4 @@
 import subprocess
-import time
 
 import pytest
 
@@ -7,7 +6,7 @@ from riotctrl_shell.gnrc import GNRCICMPv6Echo, GNRCPktbufStats
 from riotctrl_shell.netif import Ifconfig
 
 from testutils.asyncio import wait_for_futures
-from testutils.shell import ping6, pktbuf, lladdr
+from testutils.shell import ping6, lladdr, check_pktbuf
 
 
 APP = 'examples/gnrc_networking'
@@ -40,8 +39,7 @@ def test_task01(riot_ctrl):
                 count=1000, interval=20, packet_size=0)
     assert res['stats']['packet_loss'] < 10
 
-    assert pktbuf(pinged).is_empty()
-    assert pktbuf(pinger).is_empty()
+    check_pktbuf(pinged, pinger)
 
 
 @pytest.mark.iotlab_creds
@@ -64,8 +62,7 @@ def test_task02(riot_ctrl):
                 count=1000, interval=100, packet_size=50)
     assert res['stats']['packet_loss'] < 10
 
-    assert pktbuf(pinged).is_empty()
-    assert pktbuf(pinger).is_empty()
+    check_pktbuf(pinged, pinger)
 
 
 @pytest.mark.iotlab_creds
@@ -89,8 +86,7 @@ def test_task03(riot_ctrl):
                 count=500, interval=300, packet_size=1024)
     assert res['stats']['packet_loss'] < 10
 
-    assert pktbuf(pinged).is_empty()
-    assert pktbuf(pinger).is_empty()
+    check_pktbuf(pinged, pinger)
 
 
 @pytest.mark.iotlab_creds
@@ -119,8 +115,7 @@ def test_task04(riot_ctrl):
     assert res['stats']['packet_loss'] < 10
 
     pinged.start_term()
-    assert pktbuf(pinged).is_empty()
-    assert pktbuf(pinger).is_empty()
+    check_pktbuf(pinged, pinger)
 
 
 @pytest.mark.iotlab_creds
@@ -149,8 +144,7 @@ def test_task05(riot_ctrl):
                 count=1000, interval=100, packet_size=50)
     assert res['stats']['packet_loss'] < 10
 
-    assert pktbuf(pinged).is_empty()
-    assert pktbuf(pinger).is_empty()
+    check_pktbuf(pinged, pinger)
 
 
 @pytest.mark.iotlab_creds
@@ -180,8 +174,7 @@ def test_task06(riot_ctrl):
                 count=1000, interval=100, packet_size=100)
     assert res['stats']['packet_loss'] < 10
 
-    assert pktbuf(pinged).is_empty()
-    assert pktbuf(pinger).is_empty()
+    check_pktbuf(pinged, pinger)
 
 
 @pytest.mark.local_only
@@ -204,8 +197,7 @@ def test_task07(riot_ctrl):
                 count=1000, interval=100, packet_size=50)
     assert res['stats']['packet_loss'] < 10
 
-    assert pktbuf(pinged).is_empty()
-    assert pktbuf(pinger).is_empty()
+    check_pktbuf(pinged, pinger)
 
 
 @pytest.mark.local_only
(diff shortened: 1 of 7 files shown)


miri64:
> his commit fixes some race conditions when checking that all pinged nodes

Not really a race condition is it? The applications just does not wait for the stack to finish.

jia200x:
> https://github.com/RIOT-OS/Release-Specs/pull/209

But sometimes passes, sometimes not. As it was feels like a race condition to me :) (on the test side)

04-single-hop-6lowpan-icmp/test_spec04.py
```diff
@@ -12,13 +12,33 @@
...
+
+
+def check_pktbuf(nodes):
+    """ Probe every PKTBUF_PROBE_DELAY seconds if the pktbuf of all nodes
+        is empty. If so, this function returns True.
+        After PKTBUF_PROBE_ATTEMPTS failed attempts, this function
+        returns False.
+    """
+    for _ in range(PKTBUF_PROBE_ATTEMPTS):
+        if not False in [_probe_pktbuf(n) for n in nodes]:
```
miri64, line 36: Why the indirection to `_probe_pktbuf`? Just because of the `print`? Ideally this is not needed.

```suggestion
        if not False in [pktbuf(n).is_empty() for n in nodes]:
```

miri64 (changes requested):
Why does this "race condition" not cause problems in specs 3, 5, 6 and 7 where the same pattern is used to check the packet buffer? If it does, `check_pktbuf` should probably be moved to `testutils/shell.py` (and this PR should only contain the bug fix for 4.13).

miri64:
Also where do 5 and 12 come from? Why not just wait 60 sec as we do currently? Usually it is known why a packet snip stays in the packet buffer (e.g. waiting to be discarded after failed reassembly), so a test should reflect that.

miri64:
> > Not really a race condition is it? The applications just does not wait for the stack to finish.
> 
> But sometimes passes, sometimes not. As it was feels like a race condition to me :) (on the test side)

A race condition is met when an undesired race between states is happening. In this case however, the packet buffer might be empty in some cases, because all packets where sent successfully, while when it is not empty some packets might wait for a delete condition (e.g. a queuing timeout). That is not a race condition but more or less expected behavior.

jia200x:
I changed the behavior of the `check_pktbuf` function. Now it receives a timeout.

Do you have any comments regarding what should be the minimum timeout for each tests?

04-single-hop-6lowpan-icmp/test_spec04.py
```diff
@@ -12,13 +12,21 @@
...
+PKTBUF_PROBE_DELAY = 5
+PKTBUF_PROBE_ATTEMPTS = 12
 pytestmark = pytest.mark.rc_only()
 
 
 class Shell(Ifconfig, GNRCICMPv6Echo, GNRCPktbufStats):
     pass
 
 
+def check_pktbuf(nodes, timeout=60):
```
miri64, line 24: Still would prefer

```suggestion
def check_pktbuf(*nodes, timeout=60):
```

04-single-hop-6lowpan-icmp/test_spec04.py
```diff
@@ -40,8 +48,7 @@ def test_task01(riot_ctrl):
                 count=1000, interval=20, packet_size=0)
     assert res['stats']['packet_loss'] < 10
 
-    assert pktbuf(pinged).is_empty()
-    assert pktbuf(pinger).is_empty()
+    check_pktbuf((pinged, pinger))
```
miri64, line 51: and

```suggestion
    check_pktbuf(pinged, pinger)
```

in those calls.

04-single-hop-6lowpan-icmp/test_spec04.py
miri64: IMHO this function still belongs to `testutils/shell.py`.

04-single-hop-6lowpan-icmp/test_spec04.py
```diff
@@ -12,13 +12,21 @@
 
 APP = 'examples/gnrc_networking'
 TASK10_APP = 'tests/gnrc_udp'
+PKTBUF_PROBE_DELAY = 5
+PKTBUF_PROBE_ATTEMPTS = 12
```
miri64, line 16: Not needed anymore, right?

miri64 (changes requested):

miri64:
> Do you have any comments regarding what should be the minimum timeout for each tests?

The minimum timeout should be 0 (and not sleep in that case) ;-).

04-single-hop-6lowpan-icmp/test_spec04.py
```diff
@@ -12,13 +12,21 @@
...
 pytestmark = pytest.mark.rc_only()
 
 
 class Shell(Ifconfig, GNRCICMPv6Echo, GNRCPktbufStats):
     pass
 
 
+def check_pktbuf(nodes, timeout=60):
+    time.sleep(timeout)
```
miri64, line 25:
```suggestion
def check_pktbuf(nodes, wait=0):
    if wait:
        time.sleep(wait)
```

1. It's a wait time, not a timeout.
2. There are cases where one does not need to wait which should be the default.

jia200x:
> The minimum timeout should be 0 (and not sleep in that case) ;-).

I meant, what should be the wait time for each test?
For those using the pktq (e.g those with the SubMAC) is definitely not zero, since there might be packet in the queue even after sending.

miri64:
> For those using the pktq (e.g those with the SubMAC) is definitely not zero, since there might be packet in the queue even after sending.

I'd say, in the case of `pktq` take [`CONFIG_GNRC_NETIF_PKTQ_TIMER_US`](https://doc.riot-os.org/group__net__gnrc__netif__conf.html#ga2770a46f626eafdcfb5fc17ac01229a9) times number of expected fragment. Then round to the nearest multiple of 10 seconds and add another 10 if the difference between your original value and the final value is less than 5. :-)

jia200x:
I moved `check_pktbuf` to `testutils`. Could you check if everything makes sense now? I will still need to rename the commits though

miri64:
Ok, now let's see if your change in test behavior still reliably lets the tests pass again. If it works, I stop arguing about the default value for `wait`.

miri64:
If it fails, I would prefer to get back to the old values, as they had been proven to be reliable and use the same values for tasks 4.12 and 4.13, as originally planned.

jia200x:
> If it fails, I would prefer to get back to the old values, as they had been proven to be reliable and use the same values for tasks 4.12 and 4.13, as originally planned.

They passed :).
As discussed offline, in worst case a 1024 bytes IPv6 packet (11 fragments) will spend at most `CONFIG_GNRC_NETIF_PKTQ_TIMER_US` * 11 ~~ 55 ms in the queue. The timeout for reassembly is 3 seconds. Therefore, I think there's no way the pktbuf is full 4 seconds after the last fragment was sent. And so 10 seconds :)

miri64 (approved):
Then let's go with this! ACK!

miri64:
Please squash though!

jia200x:
done!

miri64:
ACK!

jia200x:
thanks for the review!

PR #211, opened 2021-02-25
99-compile-and-test-one-board: add link for 02-tests

miri64:
We have the technology, so let's link the spec `99-compile-and-test-one-board` is referring to.


kaspar030 (approved):
ACK.

PR #212, opened 2021-02-27
task8.4: fix address parsing for RIOT-OS/RIOT#15992

miri64:
Due to RIOT-OS/RIOT#15992 the parsing in [the release tests](https://github.com/RIOT-OS/RIOT/runs/1993539557) currently fails. Where

```sh
$ RIOTBASE=$(readlink -f ../RIOT) tox -- --non-RC -k "spec08 and task08"
```

failed before, it should succeed with this PR. (If you are facing the same problem I did that the test fails because of a `FileNotFoundError`: that's [a bug in python 3.9](https://bbs.archlinux.org/viewtopic.php?id=261816). Just delete `10-icmpv6-error/test_spec10.py` for the test (the tests in there are deselected with the command above anyways), and it will work.

### DIFF
--- a/08-interop/test_spec08.py
+++ b/08-interop/test_spec08.py
@@ -66,7 +66,7 @@ def test_task08(riot_ctrl):
     _, gnrc_addr = lladdr(gnrc_node.ifconfig_list())
     assert gnrc_addr.startswith("fe80::")
     res = lwip_node.cmd("ifconfig")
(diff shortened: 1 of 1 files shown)


miri64:
(I restarted the test since there was another test that failed with large ping loss, but I suspect that was just because of the medium being occupied and thus no traffic coming through on the testbed)

miri64:
Ping? Release tests are still failing due to this.

benpicco (approved):

PR #223, opened 2021-05-29
release-tests: mark IoT-LAB runs as flaky

miri64:
Oftentimes tests fail only because there is currently a lot of noise in the IoT-LAB testbed due to ongoing experimentation (as it is currently the case for some of the [latest runs](https://github.com/RIOT-OS/RIOT/actions/workflows/release-test.yml). This change allows for those tests to be repeated on failure after 30 seconds up to 3 times.

Experimental tests marked with XFAIL are skipped for this as they will be marked as skipped on failure anyway and that should not be hidden.

### DIFF
--- a/04-single-hop-6lowpan-icmp/test_spec04.py
+++ b/04-single-hop-6lowpan-icmp/test_spec04.py
@@ -18,6 +18,7 @@ class Shell(Ifconfig, GNRCICMPv6Echo, GNRCPktbufStats):
     pass
 
 
+@pytest.mark.flaky(reruns=3, reruns_delay=30)
 @pytest.mark.iotlab_creds
 # nodes passed to riot_ctrl fixture
 @pytest.mark.parametrize('nodes',
@@ -42,6 +43,7 @@ def test_task01(riot_ctrl):
     check_pktbuf(pinged, pinger)
 
 
+@pytest.mark.flaky(reruns=3, reruns_delay=30)
 @pytest.mark.iotlab_creds
 # nodes passed to riot_ctrl fixture
 @pytest.mark.parametrize('nodes',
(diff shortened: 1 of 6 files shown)


miri64:
Oh yeah, and tests that only check the packet buffer state I also did not mark as flaky, as only packet loss is affected by noise ;-).

04-single-hop-6lowpan-icmp/test_spec04.py
```diff
@@ -18,6 +18,7 @@ class Shell(Ifconfig, GNRCICMPv6Echo, GNRCPktbufStats):
     pass
 
 
+@pytest.mark.flaky(reruns=3, reruns_delay=30)
```
fjmolinas, line 21: What do you think about defining some defaults for this maybe using `pytest_configure`? Feel free to dismiss though.

miri64: If you tell me, what you mean by that I might oblige. The [doc on `pytest_configure`](https://docs.pytest.org/en/6.2.x/reference.html#pytest.hookspec.pytest_configure) isn't all that helpful.

miri64:
Rebased to fix commit signature.

miri64:
I started a test run (excluding spec03 as it only has native tests) [here](https://github.com/RIOT-OS/RIOT/actions/runs/892106200).

04-single-hop-6lowpan-icmp/test_spec04.py
miri64: For posterity: @fjmolinas and I agreed out of band, that we both do not know how to use `pytest_configure` properly and that I in addition fear that hiding config there would make the code harder to understand.

fjmolinas (approved):
ACK

miri64:
> I started a test run (excluding spec03 as it only has native tests) ~[here](https://github.com/RIOT-OS/RIOT/actions/runs/892106200)~ [here](https://github.com/RIOT-OS/RIOT/actions/runs/892145390) (had a syntax error on the input variables).

Fails seem unrelated.

PR #224, opened 2021-06-03
pytest: make builds quieter

miri64:
Whenever scrolling through the `release-tests` log, I waste a lot of time scrolling through the build log which in most cases is not really interesting. Now that `QUIETER` was introduced in https://github.com/RIOT-OS/RIOT/pull/16513 we can use that to prevent that. If for any reason, we want to check the modules used, we still can use `make info-modules` (which would have the added benefit that it also lists pseudo-modules).

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -281,6 +281,7 @@ def nodes(local, request, boards, iotlab_site):
 
 
 def update_env(node, modules=None, cflags=None, port=None, termflags=None):
+    node.env['QUIETER'] = '1'
     if not isinstance(modules, str) and \
        isinstance(modules, Iterable):
         node.env['USEMODULE'] = ' '.join(str(m) for m in modules)


fjmolinas (approved):
ACK

jeandudey (approved):

miri64:
Quick test run started over at https://github.com/RIOT-OS/RIOT/actions/runs/903351779.

miri64:
Mhhh maybe we should have waited with the merge before that run was complete... so let's cross our fingers that it works ;-)

miri64:
:tada: Task 1.2 passed, but `tasks (iotlab_creds)` failed. However, this is due to all tasks being deselected in that one, so all good :-).

miri64:
Mhhh we should have had a look at the log output in the artifacts as well, because the builds are not quieter...

Issue #227, opened 2021-07-02
Release 2021.07 - RC1


MrKevinWeiss:
This issue lists the status of all tests for the [Release Candidate 1 of the 2021.07](https://github.com/RIOT-OS/RIOT/releases/tag/2021.07-RC1) release.

Specs tested:
- [ ] [01-ci](https://github.com/RIOT-OS/Release-Specs/blob/394705f179d651f44d133d21832a982df3ed1dad/01-ci/README.md)
  - [ ] [Task #01 - Compile test](https://github.com/RIOT-OS/Release-Specs/blob/394705f179d651f44d133d21832a982df3ed1dad/01-ci/README.md#task-01---compile-test)
  - [x] [Task #02 - Unittests on native](https://github.com/RIOT-OS/Release-Specs/blob/394705f179d651f44d133d21832a982df3ed1dad/01-ci/README.md#task-02---unittests-on-native) @riot-ci (see https://github.com/RIOT-OS/Release-Specs/issues/227#issuecomment-874039214) 
  - [x] [Task #03 - Unittests on native separated](https://github.com/RIOT-OS/Release-Specs/blob/394705f179d651f44d133d21832a982df3ed1dad/01-ci/README.md#task-03---unittests-on-native-separated) @riot-ci (see https://github.com/RIOT-OS/Release-Specs/issues/227#issuecomment-874039336) 
  - [x] [Task #04 - Unittests on iotlab-m3](https://github.com/RIOT-OS/Release-Specs/blob/394705f179d651f44d133d21832a982df3ed1dad/01-ci/README.md#task-04---unittests-on-iotlab-m3) @riot-ci (see https://github.com/RIOT-OS/Release-Specs/issues/227#issuecomment-874039893) 
- [ ] [02-tests](https://github.com/RIOT-OS/Release-Specs/blob/394705f179d651f44d133d21832a982df3ed1dad/02-tests/README.md)
  - [ ] [Task #01 - All tests on native](https://github.com/RIOT-OS/Release-Specs/blob/394705f179d651f44d133d21832a982df3ed1dad/02-tests/README.md#task-01---all-tests-on-native)
  - [ ] [Task #02 - Subset of tests on iotlab-m3](https://github.com/RIOT-OS/Release-Specs/blob/394705f179d651f44d133d21832a982df3ed1dad/02-tests/README.md#task-02---subset-of-tests-on-iotlab-m3)
[...]

PR #228, opened 2021-07-05
testutils.github: use sticky comments

miri64:
This makes Murdock less spammy in the release candidate test tracking issues.

See ~~https://github.com/RIOT-OS/Release-Specs/issues/227#issuecomment-874236990~~ https://github.com/RIOT-OS/Release-Specs/issues/227#issuecomment-874717209 for a currently generated example comment.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 aiocoap[linkheader]>=0.4b3
+beautifulsoup4
 iotlabcli
 pygithub
 pytest
--- a/testutils/github.py
+++ b/testutils/github.py
@@ -3,13 +3,21 @@ import os
 import re
 import subprocess
 
+from bs4 import BeautifulSoup
 from github import Github, GithubException
 
 
+STICKY_COMMENT_COMMENT = "<!-- release-specs results {user} -->"
 GITHUB_DOMAIN = "github.com"
 API_URL = "https://api.%s" % GITHUB_DOMAIN
 REPO_NAME = "RIOT-OS/Release-Specs"
 GITHUBTOKEN_FILE = ".riotgithubtoken"
+OUTCOME_EMOJIS = {
+    "passed": "✔",
+    "failed": "✖",
+    "skipped": "🟡",
+    "rerun": "🔁",
+}
 
 
 logger = logging.getLogger(__name__)
@@ -37,12 +45,18 @@ def get_rc():
 
 
 def get_task(nodeid):
-    m = re.search(r"test_spec(?P<spec>\d+)\.py::test_task(?P<task>\d+)",
+    m = re.search(r"test_spec(?P<spec>\d+)\.py::test_task(?P<task>\d+)"
+                  r"(\[(?P<params>[0-9a-zA-Z_-]+)\])",
                   nodeid)
     if m is None:
         logger.warning("Can not find task in {}".format(nodeid))
         return None
-    return {k: int(v) for k, v in m.groupdict().items()}
+    return {k: int(v) if k != 'params' else v
+            for k, v in m.groupdict().items()}
+
+
+def get_user_name(github):
+    return github.get_user().login
 
 
 def get_access_token():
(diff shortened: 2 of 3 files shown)


MrKevinWeiss:
Something isn't looking right for me, stdout is showing on 01. Task 03 - Unittests on native separated [tests-gcoap] and below... Could it be a stray escape character?

miri64:
> Something isn't looking right for me, stdout is showing on 01. Task 03 - Unittests on native separated [tests-gcoap] and below... Could it be a stray escape character?

Yeah yeah, I tried to fix something different (the **SKIPPED** info output), and accidentally broke the other output in the process :sweat_smile:. I restarted the tests, the table should fix itself in the process.

miri64:
(with the last force push both issues should be fixed)

miri64:
> I restarted the tests, the table should fix itself in the process.

Done.

miri64:
Damn, there is a character limit on comments (65536 characters). So now the comment has become too long and there are no more comments left. Will try to find a solution for that.

miri64:
Task 4.1 alone apparently alone exceeds this character limit. I am a bit confused, how https://github.com/RIOT-OS/Release-Specs/issues/227#issuecomment-874040873 can be then, but I _do_ get an error message that I exceeded the character limit with this PR [here](https://github.com/RIOT-OS/Release-Specs/pull/228/files#diff-9d99c0b27cd7d383a7053ed1e03e8611ff67dafc95d69dad4a6d1b8d3c6de312R294).

MrKevinWeiss:
I am not opposed to having a summary of x tests passed, y tests failed, and z tests skipped with a dropdown containing a list of each fail test and a link to an external log dump?  Do we trust the system enough to not need to look into if a test passes?

miri64:
> I am not opposed to having a summary of x tests passed, y tests failed, and z tests skipped with a dropdown containing a list of each fail test and a link to an external log dump?

I was thinking about the same, however there are two caveats to that:

1. We can't get a live update, as we only can link to the run in the GitHub action. That one only delivers the most current line to the frontend, so older runs than when the page is loaded won't be shown.
2. We can't link to the exact location, as the GitHub action line is pretty much disjunct from the pytest output line (and I am not even sure pytest knows how many lines it outputted so far)

So basically we can just provide a link to the action (as we already do in the spammy comments) and one needs to search for the corresponding output themselves when the task is done (or use the artifact HTML for less searching). If we are fine with that, I would prefer to remove the dumps altogether (even for the error cases) as even when just dumping the error cases we run into the risk of hitting the character limit (case and point Task 4.1 which has >80000 characters alone).

miri64:
> Do we trust the system enough to not need to look into if a test passes?

I'd say yes. In the past the false negatives where more prevalent than the false positives.

miri64:
Another (more complicated) alternative could be to dump the output to a private [gist](https://gist.github.com/) and link to that. But I would prefer to do that as a follow-up.

miri64:
[Here](https://github.com/RIOT-OS/Release-Specs/issues/227#issuecomment-874717209) is how it would look like. Since of course there is no log of my tests online (another disadvantage), I added the required environment variables to link to the last release build just to show how it would look like when Murdock reports its results. For a local test, the outcome would just not be a link. I think the Gist variant [I proposed](https://github.com/RIOT-OS/Release-Specs/pull/228#issuecomment-874622224) could be a good solution for that as well.

MrKevinWeiss:
> case and point Task 4.1 which has >80000 characters in its output alone

I wonder how much of that is usable information...

MrKevinWeiss:
gist is nice, and somehow all in the github framework so image it would make implementation not horrible.

miri64:
> gist is nice, and somehow all in the github framework so image it would make implementation not horrible.

Yepp, but still would prefer to do that as a follow up, since new tests would be needed to keep the coverage threshold up and I believe that would make this PR grow out of proportion.

MrKevinWeiss:
Sure, I will need some time to look through everything.

MrKevinWeiss:
The code looks ok, ya 400+ lines are a lot but much of that is tests and writing parsers out step by step (which I prefer to one-line regex magic).

It is an improvement and I wouldn't mind trying it out for RC-2.  Is there anything else you wanted to add for this one?  (if not please squash)

miri64:
Squashed. The Gist output is in the works. But as I said: I would prefer to provide it as a follow-up.

MrKevinWeiss (approved):
Looks good, I think the follow up with Gist would be pretty nice as looking through the failures is a bit of a challenge now.

ACK!

MrKevinWeiss:
Just thinking out loud but it would also be cool to have some easy to diff output to show a comparison of tests that have failed from the last releases (as experimental sometimes means expected failure).

miri64:
> Just thinking out loud but it would also be cool to have some easy to diff output to show a comparison of tests that have failed from the last releases (as experimental sometimes means expected failure).

Mhhh this could be achieved by having a results repo instead of a gist I guess and then marking the latest build with a tag named like the release (candidate). Of course, gists are Git repos too, but since they are only identified by a hash it is quite hard to find them again when not stored somewhere (which in the current case I do with an HTML comment in the sticky comment). Maybe another future improvement. However, that would require touching the `release-test.yaml` in RIOT-OS/RIOT though as the state needs to come from the environment.

PR #229, opened 2021-07-08
testutils: github: add capability to upload test reports to Git repo

miri64:
As discussed in https://github.com/RIOT-OS/Release-Specs/pull/228#issuecomment-874622224, this PR adds the capability that test reports are uploaded into a Git repository during an RC test run and linked in the sticky comment in the tracking issue for easier evaluation. That Git repository is either identified by the two environment variables `RESULT_OUTPUT_DIR` (the local checkout of the git repository with properly configured remotes) and `RESULT_OUTPUT_URL` (a Github-based HTML-URL of the repository) or is generated on the fly for each RC over at gist.github.com.

~~Currently the tests pass, but the coverage went way down, so I still need to provide new tests for the new code segments.~~


miri64:
Ok, tests are now passing. I started a gist based run for which the results are shown in this https://github.com/RIOT-OS/Release-Specs/issues/227#issuecomment-876467093.

MrKevinWeiss (approved):
Lots of code but the output looks good and the tests are there.  ACK.  Anything else can happen in a followup PR as I would like to test this for RC2

PR #233, opened 2021-09-27
treewide: pylint adapt to C0209

fjmolinas:
Used `flynT` for most cases, other fixed manually or ignore (like cases where kwargs were passed or if statements in the formaters).


miri64:
> Added black and applied where needed.

Can we make this a separate PR? All these extra changes make the PR harder to review.

miri64:
Please rebase!

fjmolinas:
All green here @miri64

.github/workflows/self-test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install tox
+        python -m pip install tox
```
miri64, line 39: Unrelated?

miri64: (at least move this to its own PR with an explanation why this is needed)

fjmolinas: not needed, just esthetics

miri64:
Please squash!

miri64 (approved):
ACK. Running `black .` in the root dir now does not change anything.

fjmolinas:
> ACK. Running `black .` in the root dir now does not change anything.

Note that just `black .` would change strings, but this is skipped since is configured in tox, the equivalent would be `black . --skip-string-normalization`.

miri64:
> Note that just `black .` would change strings, but this is skipped since is configured in tox, the equivalent would be `black . --skip-string-normalization`.

Mhhh no, this not configured in `tox`, but `setup.cfg`. So also when you call `black .` in that directory ;-). `setup.cfg` is not just for `tox` ;-).

miri64:
See

- https://docs.python.org/3/distutils/configfile.html
- https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#what-on-earth-is-a-pyproject-toml-file

PR #218, opened 2021-04-27
11-lorawan: add tasks 0-4 

fjmolinas:
This PR adds the missing tasks for 11-lorawan.

Depends on https://github.com/RIOT-OS/RIOT/pull/16403

```
11-lorawan/test_spec11.py::test_task01[nodes0] PASSED                                                                            [ 16%]
11-lorawan/test_spec11.py::test_task02[nodes0] PASSED                                                                            [ 33%]
11-lorawan/test_spec11.py::test_task03[nodes0] PASSED                                                                            [ 50%]
11-lorawan/test_spec11.py::test_task04[nodes0] PASSED                                                                            [ 66%]
11-lorawan/test_spec11.py::test_task05[nodes0-otaa] PASSED                                                                       [ 83%]
11-lorawan/test_spec11.py::test_task06[nodes0-abp] PASSED                                                                        [100%]
```

PR #236, opened 2021-10-05
11-lorawan: fix old command call

jia200x:
## Contribution description

This PR fixes the call command to call `txtsnd`, since the send command doesn't exists anymore since https://github.com/RIOT-OS/RIOT/pull/16080

## Testing procedure

Make sure 11-lorawan/task{05-06} passes


jia200x:
I had the remove `tests_gnrc_lorawan_send_fail` because `txtsnd` doesn't indicate whether a packet couldn't be sent or not. We should add support for that and then reintroduce this function

jia200x:
there's a missing `expect` function in `riotctrl_shell`. What should this mock function do? @miri64 @fjmolinas 
Or is there any workaround?

jia200x:
> > there's a missing `expect` function in `riotctrl_shell`. What should this mock function do? @miri64 @fjmolinas Or is there any workaround?
> 
> Isnt it
> 
> https://github.com/RIOT-OS/Release-Specs/blob/d9a83ec4a3c108af6d1227049ae05b2f3d909abc/testutils/tests/test_pytest.py#L18-L20
> ?

I thought the same, but it's not there are 2 MockSpawn functions. This one is in the RIOT repo (under `dist/tools/...)

11-lorawan/test_spec11.py
```diff
@@ -39,7 +39,7 @@ def run_lw_test(node, ttn_client, iface, dev_id):
     node.ifconfig_flag(iface, "ack_req", enable=True)
```
fjmolinas, line 39: This sets an ACK to request which is also captured by PKTDUMP, so the assert FALSE below fails.

jia200x: The ACK frame is not forwarded to the upper layer. It shouldn't fail I guess

miri64:
> there's a missing `expect` function in `riotctrl_shell`. What should this mock function do? @miri64 @fjmolinas
> Or is there any workaround?

Not sure why you need an `expect` or for what. That's what the `ShellInteraction`s are for. Can you describe your issue a bit more precise?

11-lorawan/test_spec11.py
fjmolinas:
```
> PKTDUMP: data received:
~~ SNIP  0 - size:   3 byte, type: NETTYPE_UNDEF (0)
00000000  CA  8C  A8
~~ SNIP  1 - size:   9 byte, type: NETTYPE_NETIF (-1)
if_pid: 3  rssi: -32768  lqi: 0
flags: 0x0
src_l2addr: (nil)
dst_l2addr: 02
~~ PKT    -  2 snips, total size:  12 byte
```
This is what I get with the second transmission...

miri64:
Ah this is about the mock. This should be covered by the tests here, not in `riotctrl_shell`. Let me have a look.

fjmolinas:
> Not sure why you need an `expect` or for what. That's what the `ShellInteraction`s are for. Can you describe your issue a bit more precise?

But it's not using the shell interaction here, since the output to be matched is not the cmd output... since it's matching on `pkdump`...

miri64:
(if you for some reason are landing in the `MockSpawn` in `riotctrl_shell` something is wrong... using code path's shouldn't even use that module...)

11-lorawan/test_spec11.py
jia200x: that's strange... The ACK frame is simply a downlink frame with the ACK bit set, but that frame might also have information. That looks as if the server was sending more data. I will give it a look

fjmolinas: Note that I'm testing on a TTN3 with changes from #218

miri64:
> I had the remove `tests_gnrc_lorawan_send_fail` because `txtsnd` doesn't indicate whether a packet couldn't be sent or not. We should add support for that and then reintroduce this function

In the case of your test it would make sense to just return 0, but if you want to make it re-usable, having `self.expect_res = 0` in the `__init__` constructor and then returning `self.expect_ret` in the `expect()` method as in the pytest tests would be better. So

```diff
diff --git a/dist/pythonlibs/riotctrl_shell/tests/common.py b/dist/pythonlibs/riotctrl_shell/tests/common.py
index 342fb1be74..b7d6d6b689 100644
--- a/dist/pythonlibs/riotctrl_shell/tests/common.py
+++ b/dist/pythonlibs/riotctrl_shell/tests/common.py
@@ -14,6 +14,7 @@ class MockSpawn():
         # set some expected attributes
         self.before = None
         self.echo = False
+        self.expect_res = 0
 
     @property
     def last_command(self):
@@ -42,6 +43,9 @@ class MockSpawn():
         # always match on prompt with replwrap
         return 0
 
+    def expect(self, *args, **kwargs):
+        return self.expect_res
+
 
 class MockRIOTCtrl():
     """
```

PR #240, opened 2022-01-18
11-lorawan: mark task06 as local only

fjmolinas:
Mark task06 as local only since it will work with chirpstack(some versions) but not ttn. It can be removed once https://github.com/RIOT-OS/RIOT/issues/16962 is addressed. It should keep release-tests outputs clean IMO.

I'm also pushing for this because it seems that the failing test messes with the following iterations of the LoraWAN tests (something in the TTN state for that node). I'm able to fix it by resetting the node test: e.g.: `ttn-lw-cli end-devices reset release-tests eui-70b3d57ed00463e7-otaa`, but I haven't found an mqtt interface for this. I'm not sure why this happens either, but I can't investigate this further ATM.


fjmolinas:
@miri64 @jia200x thoughts?

jia200x (approved):
ACK on my side

PR #249, opened 2022-06-07
12-twr: initial import

fjmolinas:
This PR adds a spec to test `uwb-core` package integration in RIOT and run some basic automatic tests. When not run locally we use the positions of nodes on IoT-LAB, since not all are in LOS, we use values taken with Decawave PANS R2 firmware as references, these are provided as a CSV.


fjmolinas:
This will have to wait for https://github.com/RIOT-OS/Release-Specs/pull/250

MrKevinWeiss:
This would probably be good to get in by the end of the month.

fjmolinas:
> This would probably be good to get in by the end of the month.

I can dedidcate some after work time to get this in, but it requires a review in https://github.com/RIOT-OS/RIOT/pull/18178 first.

PR #256, opened 2022-08-02
self-test: test for python 3.10

miri64:
Updates the GitHub action to also test for python 3.10.

### DIFF
--- a/.github/workflows/self-test.yml
+++ b/.github/workflows/self-test.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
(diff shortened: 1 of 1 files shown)


chrysn (approved):
Yep, a lot of people will have 3.10 already, so it's good to check on that.

Thanks, approving. (AIU it can be set to required after that, which I plan to do -- if it doesn't go through, maybe you have the permissions to do that).

miri64:
> (AIU it can be set to required after that, which I plan to do -- if it doesn't go through, maybe you have the permissions to do that).

Done

PR #260, opened 2022-10-10
09-coap: use URI with `cord_ep register`

miri64:
The parameter for `cord_ep register` was changed in https://github.com/RIOT-OS/RIOT/pull/18053 to be a full URI, not just a server name + registration interface. This fixes the specs and tests to reflect that. /cc @Teufelchen1

### DIFF
--- a/09-coap/README.md
+++ b/09-coap/README.md
@@ -54,7 +54,7 @@ Start a native terminal for the RIOT cord_ep example app.
 
 Register with the RD server with the command:
 ```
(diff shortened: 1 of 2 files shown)


miri64:
To test, make sure that

```bash
RIOTBASE=$(readlink -f ../RIOT) tox -e test -- -k "(spec09)" --non-RC
```

succeeds.

maribu (approved):
ACK, trusting your testing. Thanks for tackling this!

PR #268, opened 2023-01-06
tox-envfile: deactivate for now

miri64:
This module does not support tox >= 4 and causes an error

    pluggy._manager.PluginValidationError: unknown hook 'tox_configure' in plugin
    <module 'tox_envfile' from ...>

=> deactivate for now.

### DIFF
--- a/.github/workflows/self-test.yml
+++ b/.github/workflows/self-test.yml
@@ -43,7 +43,7 @@ jobs:
       run: |
         pyenv="py$(echo "${{ matrix.python-version }}" | tr -d .)"
         cd Release-Specs
-        tox -e "test,${pyenv},flake8,pylint" -- --doctest-modules --self-test \
+        tox -e "test,flake8,pylint" -- --doctest-modules --self-test \
             --cov=testutils --cov-report term-missing \
             --cov-config=testutils/.coveragerc \
             --cov-fail-under=95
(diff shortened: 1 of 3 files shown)


miri64:
See e.g. https://github.com/RIOT-OS/RIOT/actions/runs/3810325353 and all preceding runs.

.github/workflows/self-test.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       run: |
         pyenv="py$(echo "${{ matrix.python-version }}" | tr -d .)"
         cd Release-Specs
-        tox -e "test,${pyenv},flake8,pylint" -- --doctest-modules --self-test \
+        tox -e "test,flake8,pylint" -- --doctest-modules --self-test \
```
miri64, line 46: Had to piggy-back this, to make the tests runnable again. There seems to be some regression with tox at the moment, but pinning to a specific python version does not make much sense with the release tests anyways.

maribu (approved):

MrKevinWeiss:
> There seems to be some regression with tox at the moment

I have really noticed this, almost every old package I have seems to be broken due to the tooling around it.

Maybe we should fix some of these things better (at least with Actions)

```yaml
run: >-
          pipx run --spec tox==3.27.1 tox
```

Something like this?

miri64:
> > 
> 
> I have really noticed this, almost every old package I have seems to be broken due to the tooling around it.
> 
> Maybe we should fix some of these things better (at least with Actions)
> 
> ```yaml
>         run: >-
>           pipx run --spec tox==3.27.1 tox
> ```
> 
> Something like this?

Yeah, I did this just now for all my packages as well ("`pip install tox<4`" though)... Not sure what is causing this... anything younger than 3 month or so just works. But as I said: The python version is not all that important for the release tests themselves and the self tests run just fine without the check (and still use the correct version) so the current workaround also is okay.

PR #271, opened 2023-01-13
conftest: riot_ctrl: reset after flash on IoT-LAB

miri64:
Should fix the issue observed in https://github.com/RIOT-OS/RIOT/actions/runs/3881157262/jobs/6619863473. Should we run another round? I only tested

```sh
RIOTBASE=$(readlink -f ../RIOT) tox -e test -- -k "spec04 and task01" --non-RC
```

locally.


kaspar030 (approved):
ACK.

PR #283, opened 2023-06-05
Fix paths to RIOT test applications

aabadie:


maribu (approved):

maribu:
CI is unhappy with lots of cryptic output :/

aabadie:
#284 fixes the problematic self tests but not sure this is the best solution.

aabadie:
Thanks!

PR #285, opened 2023-07-12
requirements.txt: pin pytest to <7.4 for the moment

miri64:
With this, the release tests should run again in RIOT master.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,10 @@ aiocoap[linkheader]>=0.4b3
 beautifulsoup4
 iotlabcli
 pygithub
-pytest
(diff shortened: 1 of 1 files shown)


miri64:
saclay.iot-lab.info doesn't seem to be reachable at the moment, but the `native` tests are running \o/

benpicco (approved):

miri64:
> There's an emergency electrical maintenance in the data center hosting the server of this site. It should be up again by the end of the day.

It's non-critical. The tests are selected again properly for `native` so they should also be for the IoT-LAB nodes.

miri64:
But we can give the release tests another spin tonight, maybe a good idea so short before feature freeze, given that they were not properly run for the past month or so.

PR #302, opened 2024-01-31
testutils: Avoid deprecated parameter

miri64:
The `login_or_token` parameter of the Github class [is deprecated](https://pygithub.readthedocs.io/en/stable/github.html#github.MainClass.Github). See also [here](https://github.com/RIOT-OS/RIOT/actions/runs/7725056382/job/21058488281#step:15:59). To prevent future surprises, let's port over to the recommended alternative, before it gets dropped.

### DIFF
--- a/testutils/github.py
+++ b/testutils/github.py
@@ -4,7 +4,7 @@ import re
 import urllib.parse
 
 from bs4 import BeautifulSoup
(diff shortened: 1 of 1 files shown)


MrKevinWeiss (approved):
ACK. Tested with current RC and updated my test results in the comment.

PR #314, opened 2025-01-10
Duplicate some native tests for native64

OlegHahm:
My proposal to add some tests for native64.


06-single-hop-udp/README.md
```diff
@@ -88,7 +105,7 @@ Sending UDP between two native nodes.
 <=10% packets lost on the receiving node.
 No leaks in the packet buffer (check with `shell_cmd_gnrc_pktbuf`).
 
-Task #06 - Empty UDP on iotlab-m3
+Task #07 - Empty UDP on iotlab-m3
```
miri64, line 108: No renumbering please. There is a process for this. See https://github.com/RIOT-OS/Release-Specs?tab=readme-ov-file#missing-a-test-or-wondering-why-it-got-replaced https://github.com/RIOT-OS/Release-Specs/blob/master/REPLACED.md

PR #315, opened 2025-01-21
fix: Change iotlabs DEFAULT_SITE to grenoble

MrKevinWeiss:
It seems like saclay has a bunch of broken nodes on it, maybe grenoble is a bit better maintained...

### DIFF
--- a/testutils/iotlab.py
+++ b/testutils/iotlab.py
@@ -13,7 +13,7 @@ from iotlabcli.experiment import (
 )
 
 
-DEFAULT_SITE = 'saclay'
+DEFAULT_SITE = 'grenoble'
 IOTLAB_DOMAIN = 'iot-lab.info'
 
 


MrKevinWeiss:
If saclay doesn't get fixed soon maybe we switch.

MrKevinWeiss:
Here is a real test: https://github.com/RIOT-OS/RIOT/actions/runs/12887496104

MrKevinWeiss:
uggg, don't be fooled by the green, all the tests were skipped due to:
```
Can't access IoT-LAB front-end grenoble.iot-lab.info via SSH. Use key without password or `ssh-agent` and connect once manually to accept the fingerprint
```

MrKevinWeiss:
Needs an update from the RIOT action side.

MrKevinWeiss:
OK there we go!

MrKevinWeiss:
Let's wait for this until we have proper info on some failures we have seen.

MrKevinWeiss:
OK, let's keep this in the back pocket.

PR #319, opened 2025-02-21
treewide: update links to RIOT examples

mguetschow:
As follow-up of https://github.com/RIOT-OS/RIOT/pull/21135 and https://github.com/RIOT-OS/RIOT/pull/21221

(Note that any workflow is expected to fail until the latter PR is merged)

PR #320, opened 2025-03-03
08-interop: use existing RIOT example folder for flashing

mguetschow:
Follow-up of #319. While the contiki example is indeed still called `examples/hello-world`, the test uses the RIOT build system to flash it on the board. We thus need to specify an existing RIOT application folder for flashing.


miri64:
Started a test run of 8.3 with current masters at https://github.com/RIOT-OS/RIOT/actions/runs/13631630572

miri64:
[worked](https://github.com/RIOT-OS/RIOT/actions/runs/13631630572/job/38100482324#step:13:59)

PR #339, opened 2026-07-11
99-compile-and-test-one-board: remove ref to RYOT

crasbe:
That workflow does not exist anymore, so delete it.


AnnsAnns (approved):
needs rebase tho

